PR #1, opened 2020-07-28
Use specMACS data from munich server, added all days.

d70-t:


leifdenby:
This is great @d70-t! What do you think about adding specmacs to a separate yaml-file? `specmacs.yml` maybe? That can then be included in the main `catalog.yml` file

d70-t:
Hey @leifdenby, I like the idea a lot, and actually did that before already.
In the meantime, I also added a couple of further commits to my branch, so I've now already added the location data of the HALO aircraft as well. Maybe I should have added feature branches from the beginning as otherwise this PR is maybe a little messy already?

d70-t:
So as I also added the halo-subfolder, maybe it is a good idea to move the specmacs and dropsonde data into it as well?

leifdenby:
Thanks 😄 I think this looks good for now. I will work out how to try a test that checks that all catalog entries validate. Unless you want to do that?

d70-t:
That's a great idea. But I'd be happy if you do it :-)

PR #11, opened 2020-08-18
Add entry for CORAL LIDAR at BCO

leifdenby:
Adds CORAL LIDAR (BCO) to catalog fetching data from AERIS

### DIFF
--- a/barbados/bco.yaml
+++ b/barbados/bco.yaml
@@ -7,7 +7,7 @@ sources:
     description: Observations made with the CORAL LIDAR at BCO
     driver: opendap
     args:
-      urlpath: https://observations.ipsl.fr/thredds/dodsC/EUREC4A/BARBADOS/BCO/Raman_Lidar_CORAL/lowResolution/version_{{version}}/T{{dt}}min/nc/coral_{{date.strftime("%y%m%d")}}_0002_0000_t.nc
+      urlpath: https://observations.ipsl.fr/thredds/dodsC/EUREC4A/BARBADOS/BCO/Raman_Lidar_CORAL/lowResolution/version_{{version}}/T{{dt}}min/nc/coral_{{date.strftime("%y%m%d")}}_0002_0000_{{content_type}}.nc
       auth: null
       chunks: {}
     parameters:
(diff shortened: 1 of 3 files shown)


leifdenby:
@RobertPincus and @d70-t I'm just bringing this to your attention as an example of using "parameters" in the catalog. I had a read through the [documentation](https://intake.readthedocs.io/en/latest/catalog.html#parameter-definition) (which is a bit sparse) and worked out how to use this. Should be useful for reducing the number of entries in the catalog.

With the CORAL LIDAR data for example you can override the time-resolution parameter:
```python
ds = cat.barbados.bco.CORAL_LIDAR(dt=60).to_dask()
``` 
whereas the following will use the default (`dt=10`):
```python
ds = cat.barbados.bco.CORAL_LIDAR.to_dask()
```

RobertPincus:
@leifdenby This is pretty interesting. Is there a way to interrogate the catalog to know what the parameters are, and what the acceptable values are? How does one iterate over all the possible choices?

leifdenby:
>  Is there a way for users to discover from the catalog what all the options are?

There is `.describe()`, but it's not very beautiful:

```python
In [9]: cat.barbados.bco.CORAL_LIDAR.describe()
Out[9]:
{'name': 'CORAL_LIDAR',
 'container': 'xarray',
 'plugin': ['opendap'],
 'description': 'Observations made with the CORAL LIDAR at BCO',
 'direct_access': 'forbid',
 'user_parameters': [{'name': 'version',
   'description': 'dataset version to use',
   'type': 'str',
   'default': '2020.07.27'},
  {'name': 'date',
   'description': 'date of observation',
   'type': 'datetime',
   'min': Timestamp('2020-01-09 00:00:00'),
   'max': Timestamp('2020-02-29 00:00:00'),
   'default': Timestamp('2020-01-09 00:00:00')},
  {'name': 'dt',
   'description': 'time resolution',
   'type': 'int',
   'allowed': [10, 20, 60],
   'default': 10},
  {'name': 'content_type',
   'description': 'dataset-content',
   'type': 'str',
   'allowed': ['b', 't'],
   'default': 'b'}],
 'metadata': {'maintainer': 'Leif Denby <l.c.denby[at]leeds.ac.uk>',
  'documentation': 'https://wiki.mpimet.mpg.de/doku.php?id=analysis:data:bco:ramanlidars:raman-lidar-co
  'tags': ['water-vapour profiles', 'temperature profiles']},
 'args': {'urlpath': 'https://observations.ipsl.fr/thredds/dodsC/EUREC4A/BARBADOS/BCO/Raman_Lidar_CORAL
  'auth': None,
  'chunks': {}}}
```

I guess it's also visible in the [intake GUI](https://intake.readthedocs.io/en/latest/gui.html) and [command line](https://intake.readthedocs.io/en/latest/tools.html#intake-client)

PR #13, opened 2020-08-18
Start work on Ron Brown catalog

leifdenby:
needs more work because filenames contain a timespan and I'm not sure how to template these yet


RobertPincus:
@d70-t Me and @lbariteau decided not to include the Level-1 lidar data in the general Ron Brown catalog - there's just too much of it and it seems to have little user value. If @leifdenby wants to make this part of the catalog he could do so, but should point to the data's new home at NOAA NCEI rather than the PSL server.

PR #16, opened 2020-10-02
Added v1 data from P3

RobertPincus:
Currently hosted on NOAA PSL servers. May be updated later to point to the archive.


RobertPincus:
Just FYI - these entries use the `opendap` intake plugin to access remote netcdf files.

d70-t (approved):
Looks perfect to me.

The current (and hopefully future) version of intake-xarray should use no authentication and automatic chunking as default, so the keys `auth` and `chunks` may be ommitted. On the other hand, we might opt to keep the keys such that we don't run into trouble if the defaults would be changed later on.

PR #19, opened 2020-12-02
Add data from SWIFT ocean drifters at NOAA PSL.

RobertPincus:


RobertPincus:
@d70-t Note that HALO flights are failing the automated testing of the intake catalog

swifts.yml
```diff
@@ -0,0 +1,46 @@
+plugins:
+  source:
+    - module: intake_xarray
+
+sources:
+  SWIFT_16:
```
d70-t, line 6: There is not yet a final "official" list of platform ids, but the closest to that I know of is [here](https://github.com/eurec4a/meta/blob/master/EUREC4A.yaml#L604). That list does not use underscores for the SWIFT names. We should agree on exactly one identifier per platform and I'd suggest that these names should be the swift platform identifiers.

Should we change it here or should we change it in the platform metadata repository?

d70-t (changes requested):
I think the HALO flights fail again because of the AERIS openDAP server.

swifts.yml
RobertPincus: Excellent catch, thanks

d70-t: I'm a little confused. Now it is `SWIFT-16` etc. here and `SWIFT16` [there](https://github.com/eurec4a/meta/blob/master/EUREC4A.yaml#L604). Do I get it right that it should then be with a dash everywhere?

d70-t (approved):

PR #22, opened 2020-12-10
disable opendap access to Aeris servers

d70-t:
Currently, the aeris opendap server is very often unavailable. This
patch changes the access method to that server to netcdf download via
http and local caching (simplecache). While this is generally a worse
idea then access via opendap, it is maybe better than no access at all.

This patch should be reverted when opendap access becomes more stable.

### DIFF
--- a/barbados/bco.yaml
+++ b/barbados/bco.yaml
@@ -5,11 +5,9 @@ plugins:
 sources:
   CORAL_LIDAR:
     description: Observations made with the CORAL LIDAR at BCO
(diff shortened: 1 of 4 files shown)


d70-t:
Hmm, that would be a lot better, yes! But I don't yet know about such a thing.
I would really like to see a way of specifying multiple access methods to the same dataset in general as this could also be used for referencing the data from multiple different servers and would support load balancing, local routing and fall back.

Also, currently the server is back up...

PR #24, opened 2020-12-11
move and update specMACS dataset information.

tmieslinger:
Hi, specMACS dataset information seemed odd next to HALO and P3. I moved it to be accessible via `cat.halo.specMACS`. Also, some experimental datasets with points and mesh data for a case study on February 5 are combined in `cat.halo.specMACS.experimental`. Further, I updated the keys to SWIR cloudmask data to the respective `flight_id` and moved them to a seperate file. So they are reachable via `cat.halo.specMACS.cloudmaskSWIR[flight_id]`


d70-t (approved):
Moving the specMACS data to HALO totally makes sense. The access per `flight_id` is also clearly superior to accessing by some name which must be created manually.

PR #25, opened 2020-12-11
change identifiers to official spelling

d70-t:
Most identifier are spelled in upper case according to out metadata
table. This PR changes
* `halo` -> `HALO`
* `p3` -> `P3`
* `bahamas` -> `BAHAMAS`
* `bahamas.ql` -> `BAHAMAS.QL`


RobertPincus:
So of course the alternative is to change the metadata table. What arguments for or against?

d70-t:
For `HALO`, `BAHAMAS` and `QL` the names are very much established already. For `P3` I can't tell.

d70-t:
One thing which would (but maybe shouldn't) be an argument for `P3` are the `flight_ids` which currently contain `P3` as prefix...

RobertPincus (approved):
If we consider the meta-data standards to be well-entrenched then this change is quite sensible in promoting uniform naming across the experiment.

PR #27, opened 2021-01-05
add a products folder and radiative profiles

tmieslinger:
I would like to add a `products` folder for additional / derived campaign datasets, such as the radiative heating profiles published in Albright et al., 2020.
Currently, the netCDF file containing the data is stored on AERIS. However, some data types are incompatible with the opendap service and the authors will update the file such that the accress via opendap will work.

Nevertheless, feel free to comment on the catalogue structure leading to the dataset: `cat.products.radiative_profiles.albright2020`

### DIFF
--- a/HALO/smart_irradiances.yaml
+++ b/HALO/smart_irradiances.yaml
@@ -56,49 +56,49 @@ sources:
     description: SMART spectral irradiances.
     driver: opendap
     args:
-      urlpath: https://observations.ipsl.fr/thredds/dodsC/EUREC4A/AIRCRAFT/HALO/SMART/2020/RF08_20220202/EUREC4A_HALO_SMART_spectral_irradiances_20200202_R1.nc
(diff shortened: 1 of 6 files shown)


RobertPincus:
@tmieslinger This seems like a great idea - do you plan to finalize the pull request? 

I would advocate for descriptive terms that don't rely on people knowing particular papers - `clear_sky` instead of `albright2020`. 

I would also advocate for `derived` or perhaps `calculations`. To me `products` implies manufacturing. But maybe we don't even need this level of hierarchy? Data from, say, SpecMACs, will include plenty of modeling.

tmieslinger:
Yes! I am waiting for Anna Lea to put the new rad_profiles file on AERIS with corrected data types so that the request via opendap actually works. This should be done any time soon. 
I agree with your comments and will use `clear-sky` and put the paper reference in the description. Concerning the hierarchy, I like the idea of less levels. `products` came form the AERIS data server structure... I will change it to `cat.radiative_profiles.clear_sky`

RobertPincus:
Excellent. I look forward to having the data available in the catalog. Thanks for doing this work, @tmieslinger

d70-t (approved):
🎉

PR #28, opened 2021-01-14
use yaml_file_cat as driver for nested catalogs

d70-t:
This is the official name from the [intake plugin directory](https://intake.readthedocs.io/en/latest/plugin-directory.html?highlight=yaml_file_cat#plugin-directory).
While it is also possible to directly specify python classes, this seems
to be odd for a general catalog. In particular, I could imagine that we
want to read intake catalogs from different programming languages at
some point in time and then hard coded python module names will not be
beneficial.


RobertPincus (approved):
This looks nice and clean. Good idea.

PR #29, opened 2021-01-14
add catalog entry for the HALO WALES Lidar

tmieslinger:


HALO/wales/main.yaml
```diff
@@ -0,0 +1,35 @@
+sources:
+  cloudparameter:
+    args:
+      path: "{{CATALOG_DIR}}/wales_cloudtop.yaml"
+    description: 'WALES Lidar cloud top heights'
```
d70-t, line 5: This description doesn't seem to fit the name of the source.
```suggestion
    description: 'WALES-Lidar cloud top height, cloud optical thickness and cloud flag data.'
```
This is indicated in the `wales_cloudtop.yaml` file.

HALO/wales/main.yaml
```diff
@@ -0,0 +1,35 @@
...
+    driver: intake.catalog.local.YAMLFileCatalog
+    metadata: {}
+
+  wv:
+    args:
+      path: "{{CATALOG_DIR}}/wales_wv.yaml"
+    description: 'WALES Lidar Water Vapour Mixing Ratio in g/kg'
```
d70-t, line 12: The data in one file (HALO-0205) contains the following attributes in the variable `humidity_mixing_ratio`:
```
Attributes:
    long_name:  water vapor volume mixing ratio
    units:      ppm
```
This suggests that the data is
* the volume mixing ratio and not the mass mixing ratio
* the units are parts per million and not g/kg i.e. parts per thousand

HALO/wales/main.yaml
```diff
@@ -0,0 +1,35 @@
...
+    driver: intake.catalog.local.YAMLFileCatalog
+    metadata: {}
+
+  adepg:
+    args:
+      path: "{{CATALOG_DIR}}/wales_adepg.yaml"
+    description: 'WALES Lidar Particle Depolarisation at 532 nm'
```
d70-t, line 19: The dataset (HALO-0205) says "aerosol depolarisation at 532 nm". I am not entirely sure if the distinction between "aerosol" and "Particle" makes any difference (probably not), but still I am wondering why you chose "Particle".

d70-t (changes requested):
Thank you for adding the WALES dataset :smile: :tada:. There are only a few minor comments.

HALO/wales/main.yaml
tmieslinger: In the corresponding quicklook plots the WALES people wrote "particle" that's way I chose that term. However "aerosol" better fits their own abbreviation "adepg" - **aerosol** depolarisation green. I change it to aerosol :)

HALO/wales/main.yaml
tmieslinger: good point. I looked at a different source for information. But this seems more plausible to look at the netCDF file directly :D

HALO/wales/wales_wv.yaml
```diff
@@ -0,0 +1,95 @@
+plugins:
+  source:
+    - module: intake_xarray
+sources:
+  HALO-0122:
+    description: WALES-Lidar Water Vapour Mixing Ratio in g/kg.
```
d70-t, line 6: Thanks for updating the description in the main catalog. However, the description in this file now doesn't match anymore with the main file. Will you update it here as well?

tmieslinger: 🙈

d70-t (approved):
:tada:

PR #32, opened 2021-01-19
add SMART

JuleRadtke:
add a catalog entry for the SMART dataset

### DIFF
--- a/HALO/main.yaml
+++ b/HALO/main.yaml
@@ -20,6 +20,13 @@ sources:
     driver: yaml_file_cat
     metadata: {}
 
+  SMART:
+    args:
+      path: "{{CATALOG_DIR}}/SMART.yaml"
+    description: SMART dataset
+    driver: yaml_file_cat
+    metadata: {}
+
   WALES:
     args:
       path: "{{CATALOG_DIR}}/wales/main.yaml"
--- a/HALO/main.yaml
+++ b/HALO/main.yaml
@@ -22,7 +22,7 @@ sources:
 
   SMART:
     args:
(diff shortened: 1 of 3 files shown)


HALO/smart.yaml
```diff
@@ -0,0 +1,11 @@
+plugins:
+  source:
+    - module: intake_xarray
+
```
d70-t, line 4: This catalog doesn't contain references to `opendap` datasets, so it can be read without the use of any plugins.

```suggestion
```

d70-t (changes requested):
Looks nice in general, I'd expect only this minor change, but let's see what happens when the tests finish.

HALO/main.yaml
```diff
@@ -20,6 +20,13 @@ sources:
     driver: yaml_file_cat
     metadata: {}
 
+  SMART:
+    args:
+      path: "{{CATALOG_DIR}}/SMART.yaml"
```
tmieslinger, line 25: below, it seems that you named the file `smart.yaml`, not capitalized. Does this cause the problem? Sorry, I didn't have a closer look so far

d70-t: Good catch! This will certainly be a problem. Macs often don't case about the case of filenames, so it might not be a problem on some computers. Linux generally cares about the case, so it will be a problem on other computers.

JuleRadtke: Ah true okay lets try with both consistent.

d70-t:
YEAH! Anticapitalism rules!!!

d70-t (approved):

PR #33, opened 2021-01-21
specMACS cloudmaskSWIR v2

d70-t:
This PR updates the specMACS cloudmaskSWIR to version 2. Currently there's only one file (for flight `HALO-0202`) available, thus the PR is started as a draft and will be updated as new files become available.


tmieslinger (approved):
Looks good! Thanks for the update 
dancingbanana

PR #34, opened 2021-02-02
update swift entry to new file version `v2.1`

tmieslinger:
The swift files noted in the catalog are gone. Instead, there seem to be new files `v2.1` available on the [NOAA server](https://psl.noaa.gov/thredds/catalog/Datasets/ATOMIC/data/swift_drifters/catalog.html). I updated the intake catalog entry accordingly and hope that the tests run through again.

### DIFF
--- a/swifts.yml
+++ b/swifts.yml
@@ -8,39 +8,39 @@ sources:
     args:
       auth: null
       chunks: {}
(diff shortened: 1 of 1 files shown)


tmieslinger:
@RobertPincus do you know more about the file change? Cloud you review this pull request?

RobertPincus (approved):
@tmieslinger Thanks for catching this. The data on the ocean side are indeed being updated but the people responsible don't always know about this catalog. Your changes are good.

PR #35, opened 2021-02-08
SMART source files have been renamed 2022 -> 2020

d70-t:


d70-t:
The original files did (accidentally) contain the year 2022 in some filenames. This has been fixed on the Aeris server but in turn broke the catalog.

RobertPincus (approved):
Looks simple enough.

PR #36, opened 2021-02-09
add cloudmasks from HALO's IR imager VELOX

tmieslinger:
I would like to add the cloud mask product from HALO's IR imager VELOX. Unfortunately AERIS is unavailable at the moment. I will renew this pull request as soon as AERIS is up again and I tested the file entries.

### DIFF
--- a/HALO/main.yaml
+++ b/HALO/main.yaml
@@ -27,6 +27,13 @@ sources:
     driver: yaml_file_cat
     metadata: {}
 
+  VELOX:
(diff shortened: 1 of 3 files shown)


d70-t (approved):
Seems to be straight forward. If the checks pass, we can merge it.

tmieslinger:
@d70-t could you rerun the checks please? The access to VELOX data via the catalog works for me now.

PR #37, opened 2021-02-09
Satellite products

observingClouds:
Hi,
I thought I just add an entry about the satellite products that are available on AERIS. It's not all of them yet, since the naming of the variables is quite variable, but for now these might suffice to let me/and others create e.g. movies via opendap :)


d70-t (commented):
The change in general looks good. But exactly the naming issue maybe needs some discussion. This is a generally recurring problem within this catalog and naming things isn't easy. However, if more and more people should depend on this catalog, changing names is a really bad idea, so we need to find good names from the beginning.

First, I'd want to question if `cat.satellites.sat.GOES16_regridded` is setting the hierarchy as we want. `satellites` and `sat` seems to be redundant whereas `GOES16_regridded` seems to mix instrument and product. I would in stead suggest
`cat.satellites.GOES16.regridded` or `cat.GOES16.regridded`.

The usage of `date` and `channel` as parameters is probably fine in this case, though we didn't use parameters very much up to now. The use of `product` may be problematic though. There should be only one part which distinguishes between these kinds of products and I am wondering if this place can be kept as it is if the other products should be included later on but their file naming is incompatible to the parameters-approach. However, I am not very experienced with intake catalog parameters, so there might be another way out?

leifdenby:
I would go for `cat.satellites.GOES16.regridded` or maybe `cat.satellites.GOES16.latlongrid`? Or the description could mention what the data is regridded too. Mentioning the time resolution in the description could be helpful too.

Sounds good otherwise to me :)

tmieslinger:
Great initiative! Thanks for getting the satellites started :)
I agree with @d70-t and @leifdenby, `cat.satellites.GOES16` seems straightforward and would make it easy to add further satellite products.

d70-t:
I like @leifdenby's variant of `cat.satellites.GOES16.latlongrid` or similar, as obviously there could be other ways to regrid the dataset.

I am still wondering how to handle the `product` parameter. I currently see the following options:
* `cat.satellites.GOES16.latlongrid(product="0.5km_01min", channel=2, date="2020-02-01")`
* `cat.satellites.GOES16.latlongrid["0.5km_01min"](channel=2, date="2020-02-01")`

The first (current) one seems to be more expressive which I like a lot. On the other Hand, I see that the other file (i.e. for "2km_01min") are one file per minute. Thus I'd probably want to be able to express one of the following as well:

* `cat.satellites.GOES16.latlongrid(product="2km_01min", channel=2, date="2020-02-01T14:17")`
* `cat.satellites.GOES16.latlongrid["2km_01min"](channel=2, date="2020-02-01T14:17")`

I am unsure if the former one could be implemented in the way, intake catalogs are currently working?

However, as I am going through the files, I am wondering if those are actually the same product and if `channel` is a sensible thing for that product? Probably this one is more like what it should look like?

* `cat.satellites.GOES16.clavrx(product="2km_01min", date="2020-02-01T14:17")`
* `cat.satellites.GOES16.clavrx["2km_01min"](date="2020-02-01T14:17")`

Which then raises the question if `product` in this case is the right thing to call that variable (is it maybe something like `resolution`?) and also wether we actually need it? Probably it would be better to just get rid of it (at least for the moment) and if we later need it, we can insert a new parameter which defaults to the current behaviour. In that case, we'd probably end up with something like:

* `cat.satellites.GOES16.latlongrid(channel=2, date="2020-02-01")`
* `cat.satellites.GOES16.clavrx(date="2020-02-01T14:17")`

Which could potentially be extended in the future (if other resolutions became available) to support something like:
* `cat.satellites.GOES16.latlongrid(resolution="2km_10m", channel=2, date="2020-02-01")`

but the `resolution` parameter would stay optional and if it is left out, we'd default to what's necessary to support the current behaviour.

observingClouds:
I like @d70-t last suggestion, because the products indeed differ.
It would also be helpful, if one could define the paths according to the parameters. E.g. maybe not all of the products are on the same server or follow the same filename structure.

observingClouds:
I updated the pull request. The regraded GOES16 satellite data can now be accessed by e.g.
```
cat = open_catalog("catalog.yml")
cat.satellites.GOES16.latlongrid(channel=2, date="2020-02-10").to_dask()
```

satellites/goes16.yaml
```diff
@@ -0,0 +1,36 @@
...
+    args:
+      urlpath: https://observations.ipsl.fr/thredds/dodsC/EUREC4A/SATELLITES/GOES-E/{{resolution}}/{{date.strftime("%Y")}}/{{date.strftime("%Y_%m_%d")}}/GOES_{{'%02d' % (channel)}}_8N-18N-62W-50W_{{date.strftime("%Y%m%d")}}.nc
+      auth: null
+      chunks: {}
+    parameters:
+      resolution:
+        description: temporal and spacial resolution of gridded product
+        type: str
+        default: 0.5km_01min
+        allowed: [0.5km_01min]
```
d70-t, line 18: To keep it a little simpler, I'd skip the parameter `resolution` for now, as I currently don't see that it will become necessary. Also if it will become necessary once, we probably want to have a different format. We can add the parameter later as a non-breaking change.

```suggestion
      urlpath: https://observations.ipsl.fr/thredds/dodsC/EUREC4A/SATELLITES/GOES-E/0.5km_01min/{{date.strftime("%Y")}}/{{date.strftime("%Y_%m_%d")}}/GOES_{{'%02d' % (channel)}}_8N-18N-62W-50W_{{date.strftime("%Y%m%d")}}.nc
      auth: null
      chunks: {}
    parameters:
```

satellites/main.yaml
```diff
@@ -0,0 +1,8 @@
+
+sources:
+
+  GOES16:
```
d70-t, line 4: Just to be consistent with the other files, I'd skip the empty lines.

```suggestion
sources:
  GOES16:
```

d70-t (approved):
Seems to be fine :tada:. There are just two minor things.

satellites/goes16.yaml
observingClouds: I thought I leave it in for now, so that I don't have to make an even longer description string to mention the spatial and temporal resolution, rather let the resolution parameter speak for itself.

d70-t:
Can you merge the current `master` into this branch? This should fix the failing check and we can proceed merging the PR.

satellites/goes16.yaml
observingClouds: @d70-t you agree with me here?!

d70-t: No, as adding the `resolution` parameter creates an additional restriction for the future evolution of the catalog. As soon as we accept that parameter into the catalog, we have to assume that people are using it, even if there is only one option. If for some reason, we find out that this parameter was not a good choice, it will be hard to get rid of it.

If we omit the parameter now, and we find out later that we really need it, we can add it at any time in the future without creating an incompatible change for the uses.

But I wouldn't be too pedantic in this case and would accept the PR even with that parameter in place.

d70-t (approved):

PR #38, opened 2021-02-09
Platform tracks

observingClouds:
What about some track data? Since the data is currently on a web-server that uses compression I had to adjust the requirements a bit. The track-files I linked here are not the official released once, but they are just slightly modified (time type change, ascii characters in global attributes) to make them work with opendap. The changes are also noted in the history attribute of the specific files. An opendap compatible version is in planning and will then be published on AERIS.


requirements.txt
```diff
@@ -4,7 +4,8 @@ intake
 xarray
 zarr
 fsspec>=0.7.4
-pydap
+-e git+https://github.com/pydap/pydap#egg=pydap
+-e git+https://github.com/Pylons/webob#egg=webob
```
d70-t, line 8: I know the reason for this. But we've had discussions before about the use of not-yet-released libraries. And the general feeling was that we shouldn't rely on those, so this change is at least critical. @RobertPincus maybe has an opinion on this as well.

satellites/main.yaml
```diff
@@ -0,0 +1,8 @@
+
+sources:
+
+  sat:
+    description: Satellite data and products from different sources
```
d70-t, line 5: There have been satellite data products mixed into this PR. They should be handled separately.

d70-t (changes requested):
Thanks for the PR. 🎉 I think having a simplified and consistent way of accessing tracks of all platforms will be a great benefit!

However, there are some issues which need to be solved. The technical things are written in the respective comments.

The other issue is naming. The current hierarchy is based on `cat.<platform>.<instrument/product>`, whereas this PR suggests the use of `cat.tracks.<platform>`. I think it would be more consistent to access the tracks via `cat.<platform>.tracks`.

RobertPincus:
@observingClouds This looks terrific but I agree with all of @d70-t's comments: the naming fits better under the platform; it's better not to require non-standard libraries; and satellite data should be included separately.

satellites/main.yaml
observingClouds: Sry 🙈 I'll change the pull request.

requirements.txt
observingClouds: I agree, it's probably good to have this not merge now since the webob-issue will hopefully fixed soon. However, the last official pydap package has been published in 2017 on pypi. Many issues have been solved by now e.g. the handling of gzipped transfers. We could make a fork and publish that on pypi to get a pip package. This is probably the best solution?!

RobertPincus: Couldn't we persuade the pydap people to update the package? Open a PR on their end?

observingClouds:
I see the motivation to put these tracks to the respective platform, but I can also imagine quite some cases, where one would like to have easy access to just the tracks of all platforms. Do we have plan to connect similar variables, such that they can be searched/loaded simultaneously?
A new package has been [requested](https://github.com/pydap/pydap/issues/212).

requirements.txt
d70-t: I filed an issue at webob, if / when they are planning to do a release.

d70-t: There will be a new release of `webob` in the upcoming days.

d70-t, line 8:
```suggestion
pydap
netcdf4!=1.5.4,!=1.5.5,!=1.5.5.1 # this is due to https://github.com/Unidata/netcdf4-python/issues/1052
```

tracks.yaml
```diff
@@ -0,0 +1,25 @@
...
+    - module: intake_xarray
+
+sources:
+  HALO:
+    description: HALO track
+    driver: opendap
+    args:
+      urlpath: https://observations.mpimet.mpg.de/repository/opendap/cf207812-20fa-4c87-bf7d-f6a7f8dcbc38/entry
+      auth: null
+      chunks: {}
```
d70-t, line 12:
```suggestion
      chunks: {}
      engine: netcdf4
```

tracks.yaml
```diff
@@ -0,0 +1,25 @@
...
+    metadata:
+      tags:
+        - track
+  ATR:
+    description: ATR track
+    driver: opendap
+    args:
+      urlpath: https://observations.mpimet.mpg.de/repository/opendap/16b222a4-0ce5-42ac-95f6-3625add76175/entry
+      auth: null
+      chunks: {}
```
d70-t, line 22:
```suggestion
      chunks: {}
      engine: netcdf4
```

requirements.txt
d70-t: Now that there's `netcdf4==1.5.6` and `intake-xarray` supports opendap via netcdf, that might be a feasible way out. See comments below.

ATR/main.yaml
```diff
@@ -0,0 +1,12 @@
...
+    description: ATR track
+    driver: opendap
+    args:
+      urlpath: https://observations.mpimet.mpg.de/repository/opendap/16b222a4-0ce5-42ac-95f6-3625add76175/entry
+      auth: null
+      chunks: {}
+      engine: netcdf4
+    metadata:
+      tags:
+        - tracks
```
d70-t, line 12: ATR uses `tracks` and HALO uses `track` as a tag. I have no particular preference, but `track` seems to be consistent to the other occurrences.
```suggestion
        - track
```

d70-t (commented):
There's only one little thing left in the catalog.

I've also had a look into the data files and am wondering about the variable `flight`. I would have expected to find something like a `flight_id` in stead of an integer.

d70-t (approved):
🎉

observingClouds:
> I've also had a look into the data files and am wondering about the variable `flight`. I would have expected to find something like a `flight_id` in stead of an integer.

I wouldn't change this now, since I don't want to modify the provided files and new files should be made available soonish.

PR #41, opened 2021-02-14
added BAHAMAS subset for ESSD publication

d70-t:
I am not yet sure, which identifier should be used to access this dataset. It seems like `subset` is the distinctive name in the dataset name, but that doesn't sound like a very specific name. Any ideas?


d70-t:
After a short discussion with Marcus, I've changed the name to `PositionAttitude`.

RobertPincus (approved):
On the P-3 I think we refer to this as `navigation` but `PositionAttitude` is even more descriptive.

PR #42, opened 2021-02-16
update SMART urls to latest version of files

JuleRadtke:
new files with corrected time shift have been uploaded on Aeris

### DIFF
--- a/HALO/smart_irradiances.yaml
+++ b/HALO/smart_irradiances.yaml
@@ -3,13 +3,6 @@ plugins:
     - module: intake_xarray
 
 sources:
-  HALO-0119:
-    description: SMART spectral irradiances.
-    driver: opendap
-    args:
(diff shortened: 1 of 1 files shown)


d70-t (commented):
Looks reasonable, but what's about flights `HALO-0119` and `HALO-0118`?

JuleRadtke:
Ja I wondered about them as well, no updated versions are on aeris for these two. I noted down to ask Kevin about it.

d70-t:
The issue with that situation is that now the format of the dataset returned by
```python
cat.HALO.SMART.spectral_irradiances["HALO-0205"].to_dask()
```
is different from the dataset returned by
```python
cat.HALO.SMART.spectral_irradiances["HALO-0119"].to_dask()
```
which makes using it a lot harder than it should be. I'd think that either there should be additional datasets for the first two flights, or the datasets should be removed from the catalog.

JuleRadtke:
Ja, true. I ask Kevin, if he will upload new versions for the two as well. Until that I delete them from the cataloque.

d70-t (approved):

PR #43, opened 2021-02-16
add entry for BACARDI

JuleRadtke:
new cataloque entry for BACARDI dataset

### DIFF
--- a/HALO/main.yaml
+++ b/HALO/main.yaml
@@ -48,6 +48,13 @@ sources:
     driver: yaml_file_cat
     metadata: {}
 
+  BACARDI:
+    args:
+      path: "{{CATALOG_DIR}}/BACARDI.yaml"
+    description: 'Broadband radiometers'
+    driver: yaml_file_cat
+    metadata: {}
+
   track:
     args:
       urlpath: https://observations.mpimet.mpg.de/repository/opendap/cf207812-20fa-4c87-bf7d-f6a7f8dcbc38/entry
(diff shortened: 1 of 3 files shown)


d70-t:
Thanks for the PR 🎉 

What is the reason to use `thredds-x.ipsl.fr` in stead of `observations.ipsl.fr` as the host in the opendap URLs? They both refer to the same host:

```
% host observations.ipsl.fr
observations.ipsl.fr is an alias for perceval.ipsl.polytechnique.fr.
perceval.ipsl.polytechnique.fr has address 129.104.53.10
% host thredds-x.ipsl.fr
thredds-x.ipsl.fr is an alias for perceval.ipsl.polytechnique.fr.
perceval.ipsl.polytechnique.fr has address 129.104.53.10
```

If there's a reason to change to `thredds-x`, then we should probably to that for all datasets on Aeris. Otherwise, I'd recommend to use `observations` also for this PR.

JuleRadtke:
Ah I was wondering about that, I just copied the opendap data url. But if it's better to be consistent, I change it!

d70-t:
There's an error with the [filename casing](https://github.com/eurec4a/eurec4a-intake/pull/43/checks?check_run_id=1910965173#step:5:87).

HALO/main.yaml
```diff
@@ -50,7 +50,7 @@ sources:
 
   BACARDI:
     args:
-      path: "{{CATALOG_DIR}}/BACARDI.yaml"
+      path: "{{CATALOG_DIR}}/smart.yaml"
```
d70-t, line 53: 😬

JuleRadtke: Oh 🙈

d70-t (approved):

PR #46, opened 2021-02-17
changed intake requirement to intake[dataframe]

d70-t:
Since the newly released version of intake 0.6.1, `to_dask()` doesn't
work anymore without this install secion. This is most likely the
current primary reason for #45 to fail.


RobertPincus (approved):
The change works, as the CI shows, and I take your word for it that this is an improvement. Maybe add a line of documentation?

PR #47, opened 2021-02-18
Velox and KT19 updated the version and deleted source files

d70-t:


RobertPincus (approved):
Seems fine. On the P-3 side we have cloud indexes from radar and IR but not necessarily cloud masks - would it be helpful to make them?

d70-t:
Thanks @RobertPincus. Vero is working on cloud masks which should go into a HALO Data overview paper in ESSD. We are planning to prepare a comparison of cloud detection attempts as a chapter within the how to eurec4a and are currently collecting the datasets into intake. I don't have a final overview about how exactly the cloud masks will be defined for the profiling instruments, but they thought a lot about harmonizing these things.

I think it will be very helpful to do the same for P-3, but if there are new products to be created, I'd recomment to wait a few days to see how the current datasets look like.

PR #48, opened 2021-02-22
update WALES cloud parameter file to V1.1

tmieslinger:
Martin Wirth put an updated version V1.1 on AERIS that is CF conform and includes slight improvements in the variables and attributes.


d70-t (approved):
Looks perfekt 🎉

PR #51, opened 2021-02-24
Hamp cloudmasks v0.9

MeraX:
This is the first try to include HAMP Radar and Radiometer cloud masks. As the mask has the same time grid as the unified dataset and was mainly derived from the uniform products, I though it fits the Uniform catalog. 

I chose the names: `HAMP_MWR_cloud_mask` and `HAMP_Radar_cloud_mask` which are analogous to the respective filenames (converting on `-` to `_`).
I was not 100% sure about the names. The underscores makes them easier to read and the MWR includes "microwave". But both thinks makes the name different from the parent datasets (`HAMPradar` and `HAMPradiometer`). Any opinions on this?

And is it okay to move all unified ymal files into their own directory?


HALO/unified/main.yaml
```diff
@@ -30,3 +30,24 @@ sources:
     description: BAHAMAS data in HALO UNIFIED dataset.
     driver: yaml_file_cat
     metadata: {}
+
+  HAMP_MWR_cloud_mask:
```
d70-t, line 34:
```suggestion
  HAMPradiometer_cloudmask:
```

HALO/unified/main.yaml
```diff
@@ -30,3 +30,24 @@ sources:
...
+    args:
+      path: "{{CATALOG_DIR}}/MWR_cloud_mask_unified.yaml"
+    description: HAMP Microwave Radiometer Cloud Mask with unified time coordinate.
+    driver: yaml_file_cat
+    metadata: {}
+
+  HAMP_Radar_cloud_mask:
```
d70-t, line 41:
```suggestion
  HAMPradar_cloudmask:
```

HALO/unified/main.yaml
```diff
@@ -30,3 +30,24 @@ sources:
...
+    args:
+      path: "{{CATALOG_DIR}}/Radar_cloud_mask_unified.yaml"
+    description: HAMP Cloud Radar Cloud Mask with unified time coordinate.
+    driver: yaml_file_cat
+    metadata: {}
+
+  HAMP_MWR_retrievals:
```
d70-t, line 48:
```suggestion
  HAMPradiometer_retrievals:
```

d70-t (changes requested):
Thanks @MeraX for this PR 👍 
Moving the unified catalogs into a separate directory is perfect. And the overall structure seems to be perfectly fine.
I can understand your reasoning behind the identifiers, but I would strongly prefer to keep the names as consistent as possible and as `HAMPradar` and `HAMPradiometer` are already in the catalog, I would strongly vote to keep those names. If you make a very strong argument to change them, we should probably change them everywhere, but as that would be a breaking change, I am hesitant to propose that.
The same goes for `cloudmask`. In previous datasets (i.e. specMACS, VELOX), `cloudmask` is written without an underscore, so this should be the same.

MeraX:
I think your changes are fine. Thanks.

d70-t (approved):

PR #56, opened 2021-03-12
BAHAMAS PositionAttitude update 1.0 -> 1.1

d70-t:


HALO/BAHAMAS_PositionAttitude.yaml
```diff
@@ -6,118 +6,118 @@ sources:
...
   HALO-0128:
     driver: opendap
     args:
-      urlpath: https://observations.ipsl.fr/thredds/dodsC/EUREC4A/AIRCRAFT/HALO/BAHAMAS/data/ESSD_HALO_BAHAMAS_Subset/EUREC4A_HALO_BAHAMAS_PositionAttitude_20200128_v1.0.nc
+      urlpath: https://observations.ipsl.fr/thredds/dodsC/EUREC4A/AIRCRAFT/HALO/BAHAMAS/data/ESSD_HALO_BAHAMAS_Subset/EUREC4A_HALO_BAHAMAS_PositionAttitude_20200128_v1.1.nc
       auth: null
       chunks: {}
 
-  HALO-0130:
+  HALO-01.1:
```
MeraX, line 55: Was this change intended?

d70-t: argh... using regexes when you don't want to do that is not a good idea...

PR #60, opened 2021-03-18
OPeNDAP: change chunk from {} to null

d70-t:
There's an issue with the interworkings of dask and OPeNDAP: If dask
decides to do chunks on a dataset, it will request the whole chunk from
the backend storage, even if only a part of it is needed in the end.
However for OPeNDAP it is best to request only the required parts from
the backend. This change disables initial dask-chunking for all datasets
which previously used only a single chunk.

While this likely will improve performance for a lot of cases, it may
not always be the best solution. If multiple datasets are to be
concatenated, dask has to be used. However, in those cases, the user
maybe knows about chunking an will define those manually.

### DIFF
--- a/ATR/main.yaml
+++ b/ATR/main.yaml
@@ -9,7 +9,7 @@ sources:
     args:
       urlpath: https://observations.ipsl.fr/thredds/dodsC/EUREC4A/PRODUCTS/TRACKS/EUREC4A_tracks_ATR_v1.0.nc
       auth: null
-      chunks: {}
+      chunks: null
       engine: netcdf4
     metadata:
       tags:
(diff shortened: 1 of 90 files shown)


RobertPincus:
@d70-t Many tests are failing - do we know why?

d70-t:
Yes, we do:

Bildschirmfoto 2021-03-18 um 16 56 14

The OPeNDAP-Server is responding with `503 Service Unavailable`. The plot shows the availability during the last 14 days.

RobertPincus:
schade

d70-t:
@RobertPincus now it passed 🎉

RobertPincus (approved):
Yippee. This looks excellent.

Issue #62, opened 2021-04-15
Name collision for SWIFT buoys


RobertPincus:
There are currently two sources of information for the six SWIFT buoys. I originally introduced a directory for the complete data store, so that the data would be accessed e.g. through `cat.swifts.SWIFT16`. Commit `574fdd2` introduced track data (from Bjorn?) for many (?) platforms; these are available as `cat.SWIFT16`.

Seems like it would be better to homogenize these in some way. Looking ahead we may want to plan for distinguishing the track data from more complete datasets.

RobertPincus:
@observingClouds I would be interested what you think.

d70-t:
The problem here has been discussed a bit in #59. Using the previous scheme of accessing "all" swift data through `cat.swifts.SWIFT16` leads to the situation where it is not possible to add more datasets to the `SWIFT16` platform. If in stead the data would be accessed using `cat.SWIFT16.all`, it would be compatible with `cat.SWIFT16.track` and the like, and thus more open to further advancements of the data collection. So that is what I would currently have on my mind.

d70-t:
I am not sure if the intake catalog hierarchy is a good place to distinguish between things. The hierarchy on its own carries very little information and it will always be subjective (at least partially) as there is no way of defining the one true best hierarchy.

RobertPincus:
@d70-t Agreed that the intake catalog hierarchy doesn't carry much information, but we want to make it easy for users to predict where data are likely to be found. 

For the SWIFTS I am happy to adopt the idea of `track` and `all`. I might add the measurement frequency if this also distinguishes them. 

A second question would be whether to treat each SWIFT as an independent platform, as the tracks currently do, or to treat them as a collection, as the complete data currently do. I don't have a strong opinion (maybe @observingClouds or @tmieslinger do) and will adopt whichever idea makes sense - but adopting either convention will require a change.

d70-t:
The current idea is that there is a **platform** hierarchy-level where each platform is parallel to each other. In my mind, if something can move independently, this qualifies as a platform, which would be a hint to do it as it stands now.

---

However, I am feeling less and less confident, if it is valuable to _formally_ distinguish between the concepts of **platform** and **instrument** on a data structure level. And stating that each instrument belongs to a platform. This creates a couple of problems which are not visible at first sight, but become more and more problematic as we go on. Some examples would be:

* Using the reasoning from the top, each dropsonde and AXBT should be their own platform.
[...]

PR #63, opened 2021-04-20
SWIFT buoys treated as individual platforms.

RobertPincus:
This reorganizes data from the SWIFT drifting buoys to address #62  

At present there are two sources of SWIFT data: `cat.SWIFTNN.track` points to the track data assembled by Bjorn and `cat.swifts.SWIFTNN` points to the complete data including the track, as produced by the data providers. 

This PR adds `cat.SWIFTNN.all` to point to All The Data and removes `cat.swifts.SWIFTNN`. 

I am not convinced, though, that it makes sense to have a `.track` entry with just the track data and `.all` for the complete data, which include the track data. Would it make more sense to simply have `cat.SWIFTNN` point directly to all the data?


RobertPincus:
@observingClouds I would be curious for your ideas.

observingClouds:
I think this is the best solution for now until we have overcome some of the hierarchical access questions by e.g. specific functions or/and search functionality.

The only issue I still see here is the accessor-name `all`. This is very inclusive and basically questions every other dataset that is already connected with a SWIFT buoy ( `track`) or any additional data that might still come. Can we come up with a different name? My best suggestion is `Level_1`, although it is only a bit more satisfying than `all`, especially because I cannot recall that we have defined what each level means in this community.

observingClouds:
> Would it make more sense to simply have cat.SWIFTNN point directly to all the data?

Not sure, if I understand your question @RobertPincus correctly. Are you proposing to bann the `track` data? I think it will be good on the long run, if we don't have `cat.<thing>` directly point to a dataset, but always have an additional accessor to be able to add further datasets without changing the access structure.

d70-t (approved):
The change looks fine. But I am with @observingClouds that `all` is probably a bit too inclusive. On the other hand, that is also how the dataset is called in its filename and if that's really all, then why not?

PR #65, opened 2021-04-22
merian Wband yaml files to check

ClauClouds:
I prepared the main.yaml and Wband.yaml files to be included for having also Wband radar data from merian on the eurec4a intake. Please check if it works/has problems :)

### DIFF
--- a/MS-Merian/FMCW94_RPG.yaml
+++ b/MS-Merian/FMCW94_RPG.yaml
@@ -1,0 +1,20 @@
+plugins:
+  source:
+    - module: intake_xarray
+
+sources:
+  motion_corrected:
+    description: daily ship motion corrected Wband radar data
+    driver: opendap
+    args:
+      urlpath:  https://observations.ipsl.fr/thredds/dodsC/EUREC4A/SHIPS/RV-MARIASMERIAN/wband_radar/corrected_from_ship_motions/eurec4a_intake/{{date.strftime("%Y%m%d")}}_wband_radar_msm_eurec4a_intake.nc
+      auth: null
+      chunks: null
+      engine: netcdf4
+    parameters:
+      date:
+        description: 'daily observations from wband radar and its weather station'
+        default: '2020-01-19 20:00'
+        type: datetime
+        min: '2020-01-19 00:00'
+        max: '2020-02-19 14:00'
--- a/MS-Merian/main.yaml
+++ b/MS-Merian/main.yaml
@@ -14,3 +14,10 @@ sources:
     metadata:
       tags:
       - track
+
+  FMCW94_RPG:
+    args:
+      path: "{{CATALOG_DIR}}/FMCW94_RPG.yaml"
+    description: 'W-band radar data'
+    driver: yaml_file_cat
+    metadata: {}


Merian/Wband.yaml
```diff
@@ -0,0 +1,17 @@
+plugins:
+  source:
+    - module: intake_xarray
+
+sources:
+  daily:
+    description: daily ship motion corrected Wband radar data
+    driver: opendap
+    args:
+      urlpath:  https://observations.ipsl.fr/thredds/dodsC/EUREC4A/SHIPS/RV-MARIASMERIAN/wband_radar/corrected_from_ship_motions/eurec4a_intake/{{date.strftime("%Y%m%d")}}_wband_radar_msm_eurec4a_intake.nc
```
d70-t, line 10: I'd suggest to specify the `netcdf4` library and `chunks` explicitly, we've had severals issues with out that before, which I'd like to avoid. Explicit `auth` is only necessary for rather old versions of `intake_xarray`.

```suggestion
      urlpath:  https://observations.ipsl.fr/thredds/dodsC/EUREC4A/SHIPS/RV-MARIASMERIAN/wband_radar/corrected_from_ship_motions/eurec4a_intake/{{date.strftime("%Y%m%d")}}_wband_radar_msm_eurec4a_intake.nc
      auth: null
      chunks: null
      engine: netcdf4
```

Merian/Wband.yaml
```diff
@@ -0,0 +1,17 @@
+plugins:
+  source:
+    - module: intake_xarray
+
+sources:
+  daily:
```
d70-t, line 6: I am wondering if `daily` is the most distinctive term for that dataset. I could imagine that there could also be `daily` raw files, `daily` files of a further processing level etc... in the future. Our goal with this catalog is that identifiers which have once been issued will ideally never change again, thus, we'd like to be careful when choosing them.

I could imagine `motion_corrected` or `daily_motion_corrected` or `motion_corrected.daily` would be more specific. I tend to opt for `motion_corrected` and skip `daily` as we didn't specify the time interval for other datasets either, but maybe this one is different, as there exist daily and hourly files.

Merian/main.yaml
```diff
@@ -0,0 +1,7 @@
+sources:
+  Wband:
```
d70-t, line 2: We usually try to find a unique name for the instrument. I am wondering if there's something more specific.

d70-t (changes requested):
Wohoo! The Wband is getting live :-) 🎉 

Here are a few small comments.

Additionally to those, the `.DS_Store` files should not be committed in this repository. Finally, we'll need to get the current master branch merged into this branch and before the final merge, a rebase which remove the intermediate commits would be great.

Merian/main.yaml
ClauClouds: so, the radar official name is FMCW 94 Ghz dual polarization radar, and it is from RPG. I think the best specification would be FMCW94_RPG. Also FMCW_94 could work. what do you prefer?

ClauClouds: I personally like FMCW94_RPG more, because it identifies the company and this is important for the radar..

Merian/Wband.yaml
ClauClouds: I put motion_corrected, I think that's the best

Merian/main.yaml
ClauClouds: @d70-t ... changing the source name is enough or shall I also change the yaml file name? please let me know :)

ClauClouds: ok, changed the filename too 👍

Merian/main.yaml
```diff
@@ -0,0 +1,7 @@
+sources:
+  FMCW94_RPG:
+    args:
+      path: "{{CATALOG_DIR}}/Wband.yaml"
```
d70-t, line 4: This link is now pointing into nothing.

d70-t:
Ok, as far as I see, there's only one `.DS_Store` still left. After that, we could try to do a rebase and run the checks to clean it all up. I'll have to go in a few minutes, you can decide if
* [ ] you want to do the rebase on your own
* [ ] we should to the rebase at a different time together
* [ ] if I should do it

ClauClouds:
you do the rebase.. I have no idea how to do it :) no time to read the note :) I think I deleted the .DS.l. ahhhhh I am a bit lost here...

d70-t (approved):

d70-t:
I've done the rebase. During that process, I discovered that due to the recent addition of the `tracks` product (#59), the ship has already been creates as `MS-Merian`. In order to have only one way to reference it, I've updated your PR to use this key as well. Thus, the data will become available as:
```python
cat["MS-Merian"].FMCW94_RPG
```

PR #66, opened 2021-04-28
Add NASA saildrones

RobertPincus:
Add `all` data for Saildrones 1026, 1060, 1061. [doi:10.5067/SDRON-ATOM0](https://doi.org/10.5067/SDRON-ATOM0)


d70-t (approved):
Looks straightforward. :+1:

One thing I am wondering though is what the meaning of "leg 1" in the `description` would be. It sounds like there could be a leg 2 or something like that as well. In that case, the use of `all` as key for the dataset might be controversial...

PR #68, opened 2021-05-19
Update MWR retrieval due to recalibration

MeraX:
After discussing with Geet, Andreas and I applied the 1.06 moisture correction for the dropsondes. This impacts the MWR BT offset correction and thus the retrieval and MWR cloud mask.


d70-t (approved):
:+1: but we'll need #69 for the checks to pass.

PR #69, opened 2021-05-19
P3 flight level and remote sensing paths changed, V1.0->V1.1

d70-t:


RobertPincus (approved):
Hi, sorry, they didn't mention to me when they would update the files. Sorry too for changing the capitalization, and thanks for cleaning up after me.

PR #70, opened 2021-05-20
Update radiative heating dataset

observingClouds:
There are three versions of this dataset available at [AERIS](https://observations.ipsl.fr/aeris/eurec4a-data/PRODUCTS/RADIATIVE-PROFILES/)

So far, this catalog pointed to the oldest file (rad_profiles.nc). The new files contain a version in their filename, such that I added the `version`-parameter.

Only issue: In favour of having versions `v1.0` and `v1.2` instead of `_v1.0` and `_v1.2`, I dropped support for the former file that does not contain any version information and has been updated several times anyway.


radiative_profiles.yml
```diff
@@ -6,6 +6,13 @@ sources:
         description: Atmospheric radiative profiles from dropsondes and radiosondes published in Albright et al., 2020.
         driver: opendap
         args:
-            urlpath: https://observations.ipsl.fr/thredds/dodsC/EUREC4A/PRODUCTS/RADIATIVE-PROFILES/rad_profiles.nc
+            urlpath: https://observations.ipsl.fr/thredds/dodsC/EUREC4A/PRODUCTS/RADIATIVE-PROFILES/rad_profiles_{{version}}.nc
             auth: null
             chunks: null
+            engine: pydap
```
d70-t, line 12: I am wondering why you chose `pydap` as the engine. While this is the default, we usually had bad experience when using it.

observingClouds: I actually cannot remember. It seems to work with both engines. I've deleted the engine key.

d70-t (approved):
:+1:

Issue #26, opened 2020-12-23
New version of radiosonde dataset available


observingClouds:
Hi guys,
I'm just in the process of uploading a new version of the radiosonde dataset. This time, it is not a tar archive, but the level1 and level2 data can be directly accessed through the AERIS THREDDS server.

@leifdenby do you want to update your zarr files, or change to the AERIS THREDDS server (https://observations.ipsl.fr/thredds/catalog/EUREC4A/PRODUCTS/MERGED-MEASUREMENTS/RADIOSOUNDINGS/v3.0.0/level2/catalog.html), or even better add both sources for a better availability in case a server is down.

I make an announcement in the data-channel, when the upload is final.
Cheers!

leifdenby:
thanks @observingClouds! I was actually thinking that maybe I should remove my zarr-based mirrors from the main repository and we just use AERIS directly instead. What do you think? I'm happy to keep my zarr-based catalog available, but maybe I'll put that on a separate repository that we can link to from this main one? Maybe in `mirrors.leifdenby_zarr` or something like that? What do you think @d70-t?

observingClouds:
Well, as long as you could keep the files up-to date (and I don't see that I should reprocess them soon) and/or make sure they see which version they are using (DOI), it might actually be great to still have that resource in case AERIS is down. It would be great, if one could have several possible resources in the catalog and intake switches (semi-)automatically, but I guess this is not yet implemented? You guys probably know more.

d70-t:
I think references to Aeris should go into the catalog. However, having an active backup is also a very good idea. There is already some progress in intake/intake#557 on providing multiple locations for one dataset, but it is not done yet.

Having a mirror structure could be an addition, but I am not so sure if we really want to have that. A result of this would be that users would have to specify some form of path manually again and most likely we'll end up in having a couple of scripts passed around which only access the "mirror" tree. This can become particularly problematic if the mirror is not complete, such that some datasets will effectively work only on the main tree while others will probably only work on the mirror tree...

leifdenby:
So, in the meantime (before mirroring is available) we could just go ahead and replace the entry backed by my server with the data on AERIS? I think adding a `data_mirrors` for now might be quite nice to keep this "backup" available. Does that sound ok?

d70-t:
Puh... I really find this one hard to decide.

* mirroring is absolutely something we should have. The OPeNDAP endpoint at Aeris had an uptime of 67% during the last two weeks.
[...]

PR #71, opened 2021-07-26
Update ATOMIC URLs

xychengso:
Updated all the NCEI data URLs in the P3 folder.

### DIFF
--- a/P3/WSRA.yaml
+++ b/P3/WSRA.yaml
@@ -9,7 +9,7 @@ sources:
       auth: null
       chunks: null
       engine: netcdf4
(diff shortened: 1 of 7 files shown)


RobertPincus:
@leifdenby This PR, based on the current version of `master`, is failing because the radiosonde datasets hosted on `derby.eu ` are not able to be opened. Should we keep pointing to this data, or should we update to the "version of record" e.g. at AERIS? @d70-t your thoughts?

d70-t:
There was #26 which is not yet resolved. It ended up in a discussion about where / how to provide backup sources of this dataset.

However, as far as I can see, that referenced issue also mentions that there is a new version of the radiosonde dataset which should be included into the catalog and subsequently the old version should probably be removed anyways. I'll try to check if I can make a PR to include the newer radiosonde datasets. That will probably fix this issue as well.

d70-t (approved):
Looks fine to me.

PR #72, opened 2021-07-27
updated radiosondes to v3

d70-t:
This includes a change from denby.io to Aeris.

see #26 for some discussion about this

### DIFF
--- a/radiosondes.yml
+++ b/radiosondes.yml
@@ -1,67 +1,47 @@
+plugins:
+  source:
+    - module: intake_xarray
+
 sources:
   atalante_meteomodem:
+    driver: opendap
     args:
-      consolidated: true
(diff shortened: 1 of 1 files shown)


d70-t:
I think this is the minimal change required for the radiosondes. The issue mentioning the update (#26) is definitely overdue and currently the old version also is not available.

Probably we'll want to create more entries for other processing levels and may want to put the radiosondes at a different place in our hierarchy (the current location doesn't match our platform/product scheme too well). But those changes seem to be independent of this fixup / version upgrade.

RobertPincus (approved):
Simple and clean. I like the move to AERIS too. Agreed that we might revisit the radiosondes-in-catalog question.

Issue #75, opened 2021-09-28
Speed up tests


d70-t:
We need to find a way to speed up tests, as the catalog now (at least since #74) grows so large that re-testing everything might not be a sensible choice anymore.

Probably useful ideas:
* split tests based on data provider in order to run them in parallel and isolate errors if a provider is not reachable
* cache results of previous runs and retest only entries which have been changed since a last test run

RobertPincus:
@d70-t I like both these ideas, but I'm not sure how to implement either of them.

PR #76, opened 2021-10-19
Parallel testing

RobertPincus:
Working on #75 I wrote the tests in parallel using a matrix strategy for Github actions. Pytest uses keywords which match a pattern. I've grouped some tests together using "or." I'm sure this could be refined to better balance setup time and testing time but all the tests get done in five minutes or so. @d70-t I thought you'd be interested.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -10,6 +10,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     strategy:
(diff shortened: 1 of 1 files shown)


RobertPincus:
@d70-t We don't have squash-merging enabled for this repo. Should we change it? I don't guess it's very interesting to see all the things I tried in the main branch of this repo.

PR #80, opened 2021-11-24
JOANNE: access netCDF files directly

d70-t:
The datasets don't work nicely over OPeNDAP and accessing them directly
via netCDF seems to be the most pragmatic choice.

See also #78 and eurec4a/how_to_eurec4a#69.

### DIFF
--- a/JOANNE/main.yaml
+++ b/JOANNE/main.yaml
@@ -4,10 +4,14 @@ plugins:
 
 sources:
   level3:
-    driver: opendap
+    driver: netcdf
     description: EUREC4A/ATOMIC JOANNE Level-3 gridded dropsonde dataset
     args:
-      auth: null
       chunks: null
-      engine: netcdf4
-      urlpath: https://observations.ipsl.fr/thredds/dodsC/EUREC4A/PRODUCTS/MERGED-MEASUREMENTS/JOANNE/v2.0.0/Level_3/EUREC4A_JOANNE_Dropsonde-RD41_Level_3_v2.0.0.nc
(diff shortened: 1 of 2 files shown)


d70-t:
The failing check is unrelated. The pipeline should succeed after #81 is merged.

JOANNE/main.yaml
```diff
@@ -4,10 +4,14 @@ plugins:
 
 sources:
   level3:
-    driver: opendap
```
RobertPincus, line 7: @d70-t This seems a great approach. What's the `simplecache::` syntax?

RobertPincus (approved):

JOANNE/main.yaml
d70-t: `fsspec` has a feature called [url chaining](https://filesystem-spec.readthedocs.io/en/latest/features.html?highlight=simplecache#url-chaining), that's what makes the `::` do. Using url chaining, one can build layers of filesystems and do interesting tricks (like opening files within zipfiles stored on a remote http server).

`simplecache` is [a caching filesystem layer](https://filesystem-spec.readthedocs.io/en/latest/features.html?highlight=simplecache#caching-files-locally) which doesn't change the presentation of the contents, but keeps a copy on local disk (by default in a temporary folder which is deleted after the Python process stops). `simplecache` assumes that the content behind the referenced resource doesn't change which makes cache invalidation trivial.

So basically what the line does is: when requesting the file once within a running process, download the entire thing and keep for reuse it as long as the process is running. The change thus increases the upfront amount of data to be loaded but decreases the number of requests (it's not one to many per variable). For small files, this approach might be better in general than opendap.

PR #81, opened 2021-11-24
SWIVT update v2.1 -> v2.3

d70-t:
The old data is gone, v2.3 is reachable, but I am not certain if that's
the most current version?

### DIFF
--- a/SWIFT/SWIFT16.yaml
+++ b/SWIFT/SWIFT16.yaml
@@ -19,6 +19,6 @@ sources:
     args:
       auth: null
       chunks: null
-      urlpath: https://psl.noaa.gov/thredds/dodsC/Datasets/ATOMIC/data/swift_drifters/EUREC4A_ATOMIC_SWIFT16_All_v2.1.nc
+      urlpath: https://psl.noaa.gov/thredds/dodsC/Datasets/ATOMIC/data/swift_drifters/EUREC4A_ATOMIC_SWIFT16_All_v2.3.nc
     description: SWIFT 16, all data fields
     driver: opendap
--- a/SWIFT/SWIFT17.yaml
+++ b/SWIFT/SWIFT17.yaml
@@ -19,6 +19,6 @@ sources:
     args:
       auth: null
       chunks: null
-      urlpath: https://psl.noaa.gov/thredds/dodsC/Datasets/ATOMIC/data/swift_drifters/EUREC4A_ATOMIC_SWIFT17_All_v2.1.nc
+      urlpath: https://psl.noaa.gov/thredds/dodsC/Datasets/ATOMIC/data/swift_drifters/EUREC4A_ATOMIC_SWIFT17_All_v2.3.nc
     description: SWIFT 17, all data fields
     driver: opendap
--- a/SWIFT/SWIFT22.yaml
+++ b/SWIFT/SWIFT22.yaml
@@ -19,6 +19,6 @@ sources:
     args:
       auth: null
       chunks: null
-      urlpath: https://psl.noaa.gov/thredds/dodsC/Datasets/ATOMIC/data/swift_drifters/EUREC4A_ATOMIC_SWIFT22_All_v2.1.nc
+      urlpath: https://psl.noaa.gov/thredds/dodsC/Datasets/ATOMIC/data/swift_drifters/EUREC4A_ATOMIC_SWIFT22_All_v2.3.nc
     description: SWIFT 22, all data fields
     driver: opendap
--- a/SWIFT/SWIFT23.yaml
+++ b/SWIFT/SWIFT23.yaml
@@ -19,6 +19,6 @@ sources:
     args:
       auth: null
       chunks: null
-      urlpath: https://psl.noaa.gov/thredds/dodsC/Datasets/ATOMIC/data/swift_drifters/EUREC4A_ATOMIC_SWIFT23_All_v2.1.nc
+      urlpath: https://psl.noaa.gov/thredds/dodsC/Datasets/ATOMIC/data/swift_drifters/EUREC4A_ATOMIC_SWIFT23_All_v2.3.nc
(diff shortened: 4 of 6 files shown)


RobertPincus:
Hi, I think we would rather point to the data at NCEI (https://www.ncei.noaa.gov/thredds-ocean/catalog/psl/atomic/swift_drifters/catalog.html) but I'd like to get @lbariteau's input, since I was expecting the NCEI data to be current.

RobertPincus:
@d70-t It's holidays in the US until next week, so be patient.

lbariteau:
@RobertPincus Pointing to the NCEI repository makes more sense since we know the link will be more stable over time. Other ftp folder may change or be reorganized. Actually our PSL ftp folder structure is in the process of being reorganized right now. So posting a link that may change in 3 months doesn't make sense. We should update the v2.3 on the NCEI website instead and I believe it's actually in the pipe.

d70-t:
I've updated the links to point to NCEI at the currently available version. I want to keep the catalog in a fully operable state as much of the time as possible, thus, I'd be happy to quicky make an update to the newer version once it is available, but also prefer to have working links within the catalog at the current state.

observingClouds:
> We should update the v2.3 on the NCEI website instead and I believe it's actually in the pipe.

Looking at the NCEI server it seems like the datasets have been updated now to v2.3 (and v2.1 is no longer available). Shall we update the links to v2.3 then? I would like the CI to fail no longer because of outdated links 😬

@RobertPincus, @lbariteau any opinions/objections?

RobertPincus:
@observingClouds It would be lovely to update to version 2.3 on the NCEI server.

d70-t:
Ok, I've updated the SWIFT data, this PR should be (somwehat) ready to merge.
However, the isotope data and the waveglider data is still broken. We should come up with another one (two, three?) PRs which take care of those files.

RobertPincus:
@observingClouds @d70-t Why don't you merge this for now, and we can ask @lbariteau to look into the other outstanding data?

lbariteau:
Hi Tobias et al. Thanks for updating the SWIFT. As for the isotope and wave glider, are you saying the links are out of date or something like that? I'll take a look at it. Cheers

d70-t:
Yes, the links don't work anymore.
Hoewver, @RobertPincus already fixed the isotopes in #87, so only the wave gliders are missing currently.

RobertPincus:
Actually, @lbariteau, for the P-3 I only updated the Picarro water vapor mixing ratios. Would be great if you could add the isotope ratios.

lbariteau:
@d70-t @RobertPincus Newbie questions to the Github masters.... it seems the original eurec4a-intake has had quite a few updates since I worked on my fork version. Prior updating the new data link, I am trying to update my fork with the current version (branche?) of eurec4a-intake. I did a "git pull upstream master" command but it seems to have merge the current original with my fork instead, i.e. I have more files than the current main one. And my fork says that "This branch is 38 commits ahead of eurec4a/eurec4a-intake:master."... Confusing... 
What is the easiest way to replace my fork with the current original so I can start working from there? Tempting to delete my fork and refork but I am guessing there must me another way.
Thanks a bunch.

d70-t:
Dear @lbariteau, `git pull` is a very common command, but I don't like it too much, because it mixes two things which you might want to think of separately. It does:

* `git fetch` for downloading all the things which are on the server and not on your machine (but doesn't change your working directory), in your case from the `upstream` server (a so called "remote")
* `git merge` for merging some new things into your currently checked out branch (in this case, `git merge upstream/master`).

I usually do `fetch` and `merge` (or `rebase`) separately to be less confused.

In order to see what makes your fork (the `master` branch of your fork) 38 commits ahead of the eurec4a `master`, you can have a look at the [network](https://github.com/eurec4a/eurec4a-intake/network) graph. You'll find your recent `merge` (due to the `pull`), as well as several commits further back (in September and October) which are only in your repository but not back in the upstream. Likely these commits have been part of #74, which have been merged as 640d834 into the upstream master by @RobertPincus as as squash commit. Due to the squash (which cleans up the back and forth between the series of commits), the (merge-) relation between your branch and the upstream is not visible anymore in the git history.

One thing, which may add up to the confusion might be, that you've made your changes to your `master` branch. A more common approach would be to create a separate branch for a certain changeset within your local clone (and the fork on github).

So to clean things up, you'll likely want to go to your local clone (on your machine) and to:

**WARNING**: this assumes that you don't have any local changes in your local `master` branch which are not yet somewhere on the upstream repository. If there are such changes, these will be lost very likely.

```
git checkout master
```
to tell git that you want to modify your local `master` branch. (This is likely a no-op as you've likely checked out that one already).

Then you want to know about the current state of the `upstream`, so you'll fetch that:
```
git fetch upstream
```

Now, you can `reset` your local branch to the upstream `master`. You'll want to do it the `--hard` way, meaning that (almost) everything which is not on the upstream will be deleted:
```
git reset --hard upstream/master
```

At this point, you'll be exactly on the same spot as the current upstream. You can now push that to your fork to clean it up as well. Normally, `github` won't let you do this, because this again might delete things, thus we'll have to apply `-f`orce:

```
git push -f origin master
```

which will clean up the 38 commits ahead.

Now that everything is cleaned up, you'll want to do your changes in a new branch, e.g. `wave_glider_fixes`:

```
git checkout -b wave_glider_fixes
```

now, you can do your changes.

If they are done, you can push them separately to your fork:

```
git push -u origin wave_glider_fixes
```

This will create that branch on your fork as well and marks your local branch as tracking (`-u`), which might spare you some typing later on.

lbariteau:
Dear @d70-t , thanks a lot for your explanations and help. This helped a lot and I learn new things too. I had to create a token to be able to do the git push -f original master. Other than that all is up-to-date now on my fork. I have made the changes to the wave glider data version and added the isotope ratio to the P3. @d70-t @RobertPincus review this and let me know. All the bests

PR #84, opened 2022-04-24
ICON LES Simulations

observingClouds:
The catalog for the ICON LES simulations is still far from being finished, but more and more people start working with the output and it would be nice to point them to the simulation section of the how to eurec4a book. The articles there (PR coming soon) are based on the current state of the ICON LES intake catalog part and give already a good overview.

### DIFF
--- a/Simulations/grids.yaml
+++ b/Simulations/grids.yaml
@@ -1,0 +1,20 @@
+plugins:
+  source:
+    - module: intake_xarray
+
+sources:
+
+  EUREC4A_PR1250m_DOM01:
+    args:
+      consolidated: true
+      urlpath: https://swift.dkrz.de/v1/dkrz_948e7d4bbfbb445fbff5315fc433e36a/grids/EUREC4A_PR1250m_DOM01.zarr
+    description: ICON grid of DOM01
+    driver: zarr
+
+  EUREC4A_PR1250m_DOM02:
+    args:
+      consolidated: true
+      urlpath: https://swift.dkrz.de/v1/dkrz_948e7d4bbfbb445fbff5315fc433e36a/grids/EUREC4A_PR1250m_DOM02.zarr
+    description: ICON grid of DOM02
+    driver: zarr
+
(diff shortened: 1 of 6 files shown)


Simulations/grids.yaml
```diff
@@ -0,0 +1,20 @@
...
+    description: ICON grid of DOM01
+    driver: zarr
+
+  EUREC4A_PR1250m_DOM02:
+    args:
+      consolidated: true
+      urlpath: https://swift.dkrz.de/v1/dkrz_948e7d4bbfbb445fbff5315fc433e36a/grids/EUREC4A_PR1250m_DOM02.zarr
+    description: ICON grid of DOM02
+    driver: zarr
+
```
d70-t, line 20: I'd suggest to use the grid UUIDs as keys in this catalog. (Probably also just as an addition to the current entries). That would allow for a usage pattern similar to what has been shown in [Getting started with Monsoon 2.0]():
```python
data = cat.simulations.ICON.experiment_1.surface_DOM01.to_dask()
grid = cat.simulations.grids[data.uuidOfHGrid].to_dask()
dwg = data.merge(grid)
```
an advantage of this approach might be, that it can be easier to swap out the data and the grid will adjust iteslf automatically.

I also was wondering where to find the vertical grids which are referenced by some datasets.

d70-t (commented):
Thanks for submitting this! The PR looks quite good and should be merged quickly.

There's one thing about naming though: `experiment_1` and `experiment_2` sound quite generic. I'm currently not aware of other ICON simulations on the EUREC4A domain, so this might be just fine. But if you've got some more specific identifiers (e.g. containing LES or maybe the abc0123-type identifiers (?)) that might be more appropriate in that case.

observingClouds:
> if you've got some more specific identifiers (e.g. containing LES or maybe the abc0123-type identifiers (?)) that might be more appropriate in that case.

I don't do these `abc0123-type` identifiers, but I'll come up with something more descriptive.

Simulations/grids.yaml
observingClouds: Done with 2ff37bcc375be97d10dde09392de0ad59460d444. Vertical grids are not yet available and indexes. I'll keep that in mind for a later PR.

PR #86, opened 2022-04-25
run tests only for catalog files which have been modified by the change

d70-t:
This PR marks test cases which are based on files with changes as compared to `master` as `modified_on_branch`. That way, it's possible to run only those tests which are relevant for the current PR by running:

```bash
pytest -m modified_on_branch
```

this will save us from the many false-positive CI errors we are currently encountering.

The PR also does weekly checks of the entire catalog such that we'll still be informed if some dataset vanishes.


observingClouds (approved):
Very nice @d70-t ! I just tested this successfully locally with a test-commit. This will save us a lot of CI-time! 🚀

observingClouds (approved):

observingClouds (approved):

observingClouds (approved):

PR #88, opened 2022-05-05
wave glider and P3 isotope ratio update

lbariteau:
Update wave gliders data version. Also add P3 isotope-analzer water vapor isotope ratios


d70-t (approved):
Thanks for the PR :tada: It looks nice an clean to me :+1:

PR #91, opened 2022-05-24
Add BCO MRR data

observingClouds:
How about some MRR data from the BCO? I thought I chunk the data that is available [on AERIS](https://observations.ipsl.fr/aeris/eurec4a-data/BARBADOS/BCO/MRR/v1.0.0/), pack it into a neat zarr file and provide it via IPFS. Currently, it's only pinned locally but am happy to pin it elsewhere.


observingClouds:
Alright, I pinned it at one of the common peers we are using.

observingClouds:
🎉

PR #95, opened 2022-05-31
BCO Ka-band radar: change data location/protocol

observingClouds:
- change BCO Ka-band radar data source from AERIS OpenDAP server to DKRZ SWIFT storage
- the OpenDAP protocol is not made for this relatively large dataset and causes problems when using it in daily research


d70-t:
😬 ipfsspec v0.3.0 did not yet make it into the current CI run. Let's wait a little while and see if we get a green light once that's available.

PR #99, opened 2022-06-14
initial botany commit

fjansson:
Add catalog entry for botany simulations.

### DIFF
--- a/Simulations/DALES/botany.yaml
+++ b/Simulations/DALES/botany.yaml
@@ -1,0 +1,11 @@
+plugins:
+  source:
+    - module: intake_xarray
+
+sources:
+  dx100m:
+    args:
+      path: "{{CATALOG_DIR}}/botany_dx100m.yaml"
+    description: 'Cloud botany LES ensemble at 100m grid spacing'
+    driver: yaml_file_cat
+    metadata: {}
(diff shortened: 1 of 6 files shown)


observingClouds:
Hi @fjansson,
to merge this before intake has been released, could you compress the json file with zip and upload it also to swift? This way we circumvent the current intake bug.
Cheers!

fjansson:
Two more commits: use the zipped json instead of json to get around the intake bug, and added a yaml for the 1536x1536 simulation. So far only the parameter json is there. Then I rebased my branch on top of the new master, hope that went right.
Cheers, 
Fredrik

observingClouds:
This currently depends on solving #102.

d70-t:
Ok, now that #102 is fixed, the test is (correctly) failing as there is currently no implementation for testing the availability of `json` data sources.

* What would be the preferred way of checking if a `json` data source is available?
* Could you add the appropriate check [here](https://github.com/fjansson/eurec4a-intake/blob/403fd32614f2bcf424fae91b7df59715337674d0/tests/test_catalog.py#L46)?

observingClouds:
@fjansson please merge [my PR](https://github.com/fjansson/eurec4a-intake/pull/1) to your branch. It adds a simple test for the `json` data source. It reads the entire data source into memory which is not perfect, but for the json files that we currently have in the catalog this is totally acceptable.

We might change it to `item.discover` when we have resolved the last open points in https://github.com/intake/intake/pull/675

fjansson:
Done. Thanks!

observingClouds (approved):
This has been a little bit of an effort but now this looks very good. Congratulations everyone who contributed! 🚀 A few changes we made here will make the work also easier in the future.

observingClouds:
Time to work on https://github.com/eurec4a/how_to_eurec4a/pull/81 :smile:

PR #105, opened 2022-07-13
renamed weekly check pipeline

d70-t:
I've got the suspicion that naming this `build` causes the PRs to wait for a week to be acceptable.


observingClouds (approved):
Sounds reasonable!

PR #108, opened 2022-07-25
update SWIFT v2.3 -> v2.4

d70-t:
Files at version 2.3 disappeared from THREDDS.


RobertPincus (approved):
Looks great thanks for fixing.

PR #111, opened 2022-07-25
Remove pydap dependency

d70-t:
Pydap was causing troubles on an on and as we require netCDF anyways, we could also just use netCDF for accessing opendap resources. This PR sets all opendap resources to be accessed via netCDF.

During creation of this PR, another issue surfaced: some netCDF datasets are incompatible with the classic data model of netCDF (which is assumed to be the opendap data model by several clients). This PR changes access to the affected datasets to direct download of the original netCDF files (using `simplecache`) and opens those files as local netCDF files. I expect this to be a good trade-off, as the relevant netCDF files are rather small (several MB).

### DIFF
--- a/HALO/BAHAMAS_PositionAttitude.yaml
+++ b/HALO/BAHAMAS_PositionAttitude.yaml
@@ -10,6 +10,7 @@ sources:
       auth: null
       chunks: null
 
+      engine: netcdf4
   HALO-0118:
     driver: opendap
     args:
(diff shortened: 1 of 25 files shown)


observingClouds (approved):
Thanks for doing this elaborative, boring task! The current pydap issue should be fixed in the newest release of xarray, but as you say, not having pydap is probably more sustainable.

PR #113, opened 2022-07-25
exclude netCDF-Python v1.6.0

d70-t:
This version seems to come with curl certificate issues.
This should fix #108, #109, #110 and #111.


observingClouds (approved):
Nice catch! 🚀

PR #114, opened 2022-07-26
remove item.discover() in tests

### LINKED ISSUE #102: Remove item.discover from test
Is this line necessary? https://github.com/eurec4a/eurec4a-intake/blob/6bbfab8a59af878d7e8aedde9a4ce859da9dcedd/tests/test_catalog.py#L43

There is an issue with the `discover` method for JSON-typed catalog entries like proposed in #99 ([see failing test](https://github.com/eurec4a/eurec4a-intake/pull/99/checks#step:6:121)). The `discover`-method is currently not implemented, but I proposed a solution at https://github.com/intake/intake/issues/673

However, I do not see why we necessarily have to call `item.discover` anyway. Can we drop it or at least use an alternative?

Cheers!

d70-t:
discover is seemingly not implemented for all data sources, thus this
test can fail for the wrong reasons. Until discover is implemented for
all sources (if that happens at all) we'll have to specialize (meta-)
data loading in tests to each driver.

fixes #102

### DIFF
--- a/tests/test_catalog.py
+++ b/tests/test_catalog.py
@@ -40,7 +40,6 @@ def test_get_intake_source(catalog, dataset_name):
     if item.container == "catalog":
         item.reload()
     else:
-        item.discover()
         plugin = item.describe()['plugin'][0]
         if plugin in ["opendap", "zarr", "netcdf"]:
             _ = item.to_dask()


observingClouds (approved):

d70-t:
I didn't understand what `.discover()` does in this case, but is not covered by `.to_dask()` or `.get()`? As far as I understood, `discover` would internally call e.g. `to_dask` or `get` and extract some information, but probably we are fine if any call touching the data succeeds... And we already raise if the latter don't work.
@leifdenby could you tell me why we would need it?

PR #117, opened 2022-08-27
Add datasets of effective radii of ICON control run

observingClouds:


d70-t (approved):
Thanks for adding these datasets :+1:

Issue #121, opened 2022-09-19
Weekly tests fail due to netCDF certificate issue


observingClouds:
Just as a note, there is supposedly an [issue](https://github.com/Unidata/netcdf4-python/issues/1179) with the current netcdf4 wheel that looks like it could also cause our weekly check to fail.

I feel like this is a reoccurring error we had also a couple of months back.

d70-t:
This is a pity... I excluded that version of netCDF4 again. I relly hope the next release of netCDF4 will not have that bug. We could of course also restrict to `<1.6.0`, but then we won't get any updates anymore...

observingClouds:
Closing for now as it is fixed by #122

observingClouds:
A new version of netCDF4-python (1.6.2) has been released, but it is still causing the same troubles as described above.
According to the linked issue, an installation via conda or on a system other than linux is not causing these troubles.

For now I propose to restrict the netCDF4-python versions further in the requirements.

observingClouds:
And it's [happening again](https://github.com/eurec4a/eurec4a-intake/actions/runs/5102196921/jobs/9171601891#step:5:69868)...

observingClouds:
Fixed for now...potentially until the next netcdf version release.

PR #122, opened 2022-09-19
restrict netCDF4 even more

d70-t:
This should be a hotfix for #121. Hopefully newer releases of netCDF won't have this bug 😬.


observingClouds (approved):
Yeah, this is the best we can do right now. Maybe we want to revisit these versions at some point and check if some of the issues have been fixed.

PR #124, opened 2022-10-03
change python version to 3.8

observingClouds:
The [current failure](https://github.com/eurec4a/eurec4a-intake/actions/runs/3167336245/jobs/5157754971#step:5:37792) of the weekly test made me think that it could be a good time to finally switch to python3.8 for the tests.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -21,7 +21,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v1
       with:
-        python-version: 3.7
+        python-version: 3.8
     - name: Install dependencies (pip)
       run: |
         python -m pip install --upgrade pip
--- a/.github/workflows/weekly_tests.yml
+++ b/.github/workflows/weekly_tests.yml
@@ -14,7 +14,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v1
       with:
-        python-version: 3.7
+        python-version: 3.8
     - name: Install dependencies (pip)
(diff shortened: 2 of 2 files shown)


observingClouds:
The issue is caused by the newest release of `importlib_metadata` [5.0.0](https://github.com/python/importlib_metadata/commit/009ace37032742922d5fa04c62f04cea703ade2d) which is necessary for `python<3.8`, but does no longer support [the method used by xarray](https://github.com/pydata/xarray/blob/f5884a2070fb6cdaa9f187cfe56c3670cfad2f87/xarray/backends/plugins.py#L105)

observingClouds:
[`xarray` anyway does not support python 3.7](https://github.com/pydata/xarray/blob/f5884a2070fb6cdaa9f187cfe56c3670cfad2f87/setup.cfg#L76) since a few versions.

observingClouds:
For reference: https://github.com/pydata/xarray/issues/7115

leifdenby:
> The issue is caused by the newest release of `importlib_metadata` [5.0.0](https://github.com/python/importlib_metadata/commit/009ace37032742922d5fa04c62f04cea703ade2d) which is necessary for `python<3.8`, but does no longer support [the method used by xarray](https://github.com/pydata/xarray/blob/f5884a2070fb6cdaa9f187cfe56c3670cfad2f87/xarray/backends/plugins.py#L105)

yeah, packages are such a pain without importlib metadata. I vote that we move to requiring >= python3.8

leifdenby (approved):

PR #128, opened 2022-10-12
Add botany 1536 entries

observingClouds:
Hi @fjansson , @martinjanssens,

I added the entries for the Botany dataset (domain 1536) that are currently uploading to the catalog. Please have a look if things look as they should.

I also consolidated the uploads (aka creating a `.zmetadata` file) to make the initial load of the metadata faster.

Cheers


observingClouds:
You can check the datasets now with e.g.

```python
url = "https://raw.githubusercontent.com/observingClouds/eurec4a-intake/botany/catalog.yml"
cat = open_catalog(url)
botany_cat = cat.simulations.DALES.botany

ds_2D = botany_cat.dx100m.nx1536['2D'].to_dask()
ds_2D
```
image

observingClouds:
Here is a random snapshot:
image

observingClouds:
Alright, @martinjanssens @fjansson I think the upstream PR for the How-To-EUREC4A book is ready to be merged.
If I remember correctly you do not upload the entire `nx768` simulations for now. I guess this is reasonable as we need to see how much other simulations will actually add to the swift storage. I would suggest to remove the catalog entries for `nx768` for now. Do you agree?

martinjanssens:
Totally fine to remove these from my side for now, as `nx768` is incomplete. Do we then want to make a final update to https://github.com/eurec4a/how_to_eurec4a/pull/81 once you have merged this PR, to:
- explicitly mention that `nx1536` is the only set of simulations currently available through `eurec4a-intake`, but that more (`nx768`, as well as single points in larger cubes, like `nx3072`) might follow
- refer directly to the catalog (and not your fork+branch, @observingClouds)

observingClouds:
Sounds good to me @martinjanssens. I went ahead and removed the `nx_768` entries. I'll merge this when the test successfully finished.

Could you please go ahead and make the suggested changes to the howtoEUREC4A PR?

Simulations/DALES/botany_dx100m.yaml
```diff
@@ -3,12 +3,6 @@ plugins:
     - module: intake_xarray
 
 sources:
-  nx768:
-    args:
-      path: "{{CATALOG_DIR}}/botany_dx100m_nx768.yaml"
-    description: 'Cloud botany LES ensemble at 100m grid spacing'
```
RobertPincus, line 9: "Cloud botany" is not very informative to outsiders, so perhaps this description could be expanded.

RobertPincus (approved):
You could be a little more explicit about what the data set is but otherwise it seems fine.

PR #129, opened 2022-11-25
Exclude netCDF4 version 1.6.2

### LINKED ISSUE #121: Weekly tests fail due to netCDF certificate issue

observingClouds:
fixes #121


RobertPincus (approved):
Thanks, this is a great deal of attention to detail.

PR #133, opened 2023-02-10
ICON LES DOM03

observingClouds:
- adds output of DOM03


Simulations/grids.yaml
```diff
@@ -18,6 +18,13 @@ sources:
     description: ICON LES horizontal grid of DOM02 (EUREC4A_PR1250m_DOM02)
```
RobertPincus, line 18: This description could be made more informative for outsiders and/or a reference provided.

RobertPincus (approved):
Love the Zarr availability. Only comment is wondering if the description might be made a little more informative.

Simulations/grids.yaml
observingClouds: Thanks! I made the description a bit more explicit now.

PR #135, opened 2023-02-21
add setup for EUREC4A-MIP

observingClouds:
This PR eventually allows to load the boundary conditions for the EUREC4A-MIP experiments.

To load the hourly netCDF files into one dataset, the following steps can be done:

```python
import xarray as xr
from pandas import date_range

dates_of_interest = date_range('2020-01-01','2020-02-01', freq='1H')

cat = open_catalog("catalog.yml")
bc_cat = cat.simulations.EUREC4A_MIP.control.setup.boundary_conditions  # control
#bc_cat = cat.simulations.EUREC4A_MIP.warming.setup.boundary_conditions  # warming experiment
ds = xr.concat([bc_cat(date=d).to_dask() for d in dates_of_interest], dim='time')
ds.to_netcdf("EUREC4A-MIP_boundaryConditions.nc")
```

This code (or something equivalent) should probably make it into the howto.eurec4a book.

## To-Do:
- [x] save all files in one folder without monthly subfolders to easy download request
- [x] adapt download path in the catalog (currently only works for January datasets)

### DIFF
--- a/Simulations/EUREC4A-MIP/control/setup/bc.yaml
+++ b/Simulations/EUREC4A-MIP/control/setup/bc.yaml
@@ -5,18 +5,13 @@ plugins:
 sources:
 
   ERA5:
-    args:
-      urlpath: simplecache::swift://swift.dkrz.de/dkrz_0913c8f3-e7b6-4f94-9221-06880d4ccfea/CTRL_ERA5/cas{{date.strftime("%Y%m%d%H%M%S")}}.nc
-      chunks: null
-    parameters:
-      date:
-        description: date
-        default: 2020-01-01
-        type: datetime
-        min: 2020-01-01
-        max: 2020-03-01
+    driver: zarr
     description: Boundary conditions for storm resolving models
-    driver: netcdf
+    args:
+      urlpath: reference://
+      storage_options:
+        fo: https://swift.dkrz.de/v1/dkrz_0913c8f3-e7b6-4f94-9221-06880d4ccfea/CTRL_ERA5/CTRL_ERA5.ref
+      consolidated: false
     metadata: {}
 
   COSMO:
--- a/Simulations/EUREC4A-MIP/control/setup/bc.yaml
+++ b/Simulations/EUREC4A-MIP/control/setup/bc.yaml
@@ -1,0 +1,27 @@
+plugins:
+  source:
+    - module: intake_xarray
+
+sources:
+
+  ERA5:
+    args:
+      urlpath: simplecache::swift://swift.dkrz.de/dkrz_0913c8f3-e7b6-4f94-9221-06880d4ccfea/CTRL_ERA5/cas{{date.strftime("%Y%m%d%H%M%S")}}.nc
+      chunks: null
+    parameters:
+      date:
+        description: date
+        default: 2020-01-01
+        type: datetime
+        min: 2020-01-01
+        max: 2020-03-01
+    description: Boundary conditions for storm resolving models
+    driver: netcdf
+    metadata: {}
+
+  COSMO:
+    args:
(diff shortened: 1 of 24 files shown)


Simulations/EUREC4A-MIP/warming/setup.yaml
```diff
@@ -0,0 +1,22 @@
...
+      xarray_kwargs:
+        engine: h5netcdf
+    parameters:
+      date:
+        description: date
+        default: 2020-01-01
+        type: datetime
+        min: 2020-01-01
+        max: 2020-03-01
+    description: Boundary conditions of 4K warming experiments
```
observingClouds, line 20: @torresalavez, do I understand correctly that these PGW+4K fields can be used as is and are not just tendencies that need to be added to the boundary conditions from the control setup, right?

observingClouds: @torresalavez my assumption here is correct, right?

observingClouds: Based on the values in the datasets I assume my assumption is correct.

Simulations/EUREC4A-MIP/control/setup/bc.yaml
```diff
@@ -0,0 +1,27 @@
+plugins:
+  source:
+    - module: intake_xarray
+
+sources:
+
+  ERA5:
+    args:
+      urlpath: simplecache::swift://swift.dkrz.de/dkrz_0913c8f3-e7b6-4f94-9221-06880d4ccfea/CTRL_ERA5/cas{{date.strftime("%Y%m%d%H%M%S")}}.nc
```
d70-t, line 9: During the last years, it came more and more apparent, that large datasets are much easier to handle than small datasets (because cutting them apart is much easier than glueing them together). Especially, as the example in the pull request shows that the intended use is "glueing the things together", I'd ask if it would be more reasonable to make the data available as one dataset. In this case, the example would boild down to e.g.:

```python
import xarray as xr

cat = open_catalog("catalog.yml")
ds = cat.simulations.EUREC4A_MIP.control.setup.boundary_conditions.to_dask() \
    .sel(time=slice('2020-01-01','2020-02-01'))  # control
#ds.to_netcdf("EUREC4A-MIP_boundaryConditions.nc")
```

I could see two ways to go there:
* keep the netCDF files as they are and add a kerchunk reference file to access the dataset
* concatenate and convert the netCDF files to zarr and re-upload them

requirements.txt
```diff
@@ -9,3 +9,4 @@ s3fs
 ipfsspec
 requests
 intake-xarray>=0.3.2
+git+https://github.com/fsspec/swiftspec.git@main#egg=swiftspec
```
d70-t, line 12: When accessing the data via `https://` protocol, there would be no need anymore for `swiftspec`.
```suggestion
```

Simulations/EUREC4A-MIP/control/setup/bc.yaml
observingClouds: Yeah, I'm totally with you here @d70-t. Unfortunately, I am not the creator of the dataset and I do not want to post-process the dataset myself. This dataset is however a bit special in the sense that it will be input to climate simulations and they require mostly netCDF files which might even be allowed to only contain individual days or time steps.

That being said, I like the idea of using kerchunk references here as it would solve this dilemma. This could also be a good idea for some other datasets, e.g. the cloud radar data from the Meteor. Where would be a good place to store those index files?

observingClouds: The reference files for this case will be saved alongside the output files.

d70-t (approved):
Looks awesome :100:

PR #136, opened 2023-05-01
fix weekly tests

observingClouds:
This PR does
- add timeout to tests to limit CI usage
- add saildrone data to IPFS
- spin-up local IPFS node for tests
- change ICON-SRM and BCO radiation IPFS pins as old ones were not findable


d70-t:
`ipfsspec` v0.4 [is released](https://github.com/fsspec/ipfsspec/releases/tag/v0.4.0). Would this be an option for a more permanent fix?

d70-t:
I'm wondering why the current test fails. Line 1 of the [install_ipfs script](https://github.com/eurec4a/ipfs_tools/blob/main/install_and_run_ipfs.sh) wants to run `/bin/bash`... Does the error mean, that `/bin/bash` is not found?

observingClouds:
> `ipfsspec` v0.4 [is released](https://github.com/fsspec/ipfsspec/releases/tag/v0.4.0). Would this be an option for a more permanent fix?

That solves this fix, yes! Probably we have to exclude the newest netCDF4 library again as well. That's what I'll test next. Thanks for doing the release!

observingClouds:
> I'm wondering why the current test fails. Line 1 of the [install_ipfs script](https://github.com/eurec4a/ipfs_tools/blob/main/install_and_run_ipfs.sh) wants to run `/bin/bash`... Does the error mean, that `/bin/bash` is not found?

Seems like that was [temporarily](https://github.com/eurec4a/eurec4a-intake/actions/runs/4850741053/jobs/8644588242)

observingClouds:
Unfortunately, this [does not fix](https://github.com/eurec4a/eurec4a-intake/actions/runs/4854752472/jobs/8652496634) the weekly test entirely.

observingClouds:
@d70-t any idea how to figure out which tests fail even in this case where the CI times out and the log file is not uploaded?

observingClouds:
Now look at this. All checks have passed and even the [weekly check](https://github.com/eurec4a/eurec4a-intake/actions/runs/6401911743/job/17377699768) passes. I started a plan with Piñata and pinned the IPFS content there for now as the retrieval has worked reliably from their service so far. Time will tell and I need to find some longer term funding for this. But I would vote for merging this now. This would make a good start for version 1.0.0

d70-t (approved):
Thanks a LOT for these updates! They look great! 🎉

PR #137, opened 2023-05-01
allow manual running of weekly_test

observingClouds:
Allow manual triggering of workflow `weekly_test` because it is not triggered when testing PRs (which is good), but when fixing failures of `weekly_test` it would be necessary

see https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow


observingClouds:
I'd like to merge this and test afterwards if #136 really fixes the weekly test.

d70-t (approved):

PR #143, opened 2023-08-09
Exclude netcdf version 1.6.4 and 1.6.5

observingClouds:
continuation of #121


d70-t (approved):
Thanks very much for figuring this out!

Issue #145, opened 2023-10-09
Add .zenodo.json


d70-t:
When creating releases with a DOI, we need a bit of metadata about contributors etc...
It seems that a [.zenodo.json](https://developers.zenodo.org/#github) could help. Should we add this?

PR #147, opened 2023-10-09
Zenodo

d70-t:
This is a first shot at #145. However, the more involved parts are still missing. We should probably have:

* [x] `creators`: The list of authors, including names, affiliations, orcid
* [x] `license`: Which license do we have / want? `cc-zero` ? 
* [x] `upload_type`: e.g. `dataset` or `software`?
* [ ] `keywords` ...

[here are more options](https://developers.zenodo.org/#depositions)

### DIFF
--- a/.github/workflows/validate-metadata.yml
+++ b/.github/workflows/validate-metadata.yml
@@ -11,6 +11,6 @@ jobs:
         uses: walbo/validate-json@v1.1.0
         with:
           files: .zenodo.json
-          schema: https://zenodo.org/schemas/deposits/records/legacyrecord.json
+          schema: http://zenodo.org/schemas/deposits/records/legacyrecord.json
           fail-on-missing-schema: true
           strict: true
--- a/.github/workflows/validate-metadata.yml
+++ b/.github/workflows/validate-metadata.yml
@@ -1,6 +1,6 @@
 name: Validate JSON
 
-on: push
+on: [push, pull_request]
 
 jobs:
   validate:
--- a/.github/workflows/validate-metadata.yml
+++ b/.github/workflows/validate-metadata.yml
@@ -11,6 +11,6 @@ jobs:
         uses: walbo/validate-json@v1.1.0
         with:
           files: .zenodo.json
-          schema: http://zenodo.org/schemas/deposits/records/legacyrecord.json
+          schema: tests/json/legacyrecord.json
           fail-on-missing-schema: true
           strict: true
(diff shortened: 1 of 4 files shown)


observingClouds:
Thanks @d70-t to start working on this. So far I have always changed the metadata on the Zenodo interface after the automatic upload but the deposition file seems to be a more consistent approach.

As the deposition influences the citation format, I think we should first settle on a format that we like. I suggest
```
EUREC4A community (2023). EUREC4A [Data set]. Zenodo. 10.5281/ZENODO
```

This would imply the following parameters:
`upload_type`: `dataset`  (because we like the intake catalog to become the _EUREC4A Dataset_)
`title`: `EUREC4A` (this would result in a reference like ...EUREC4A [Data set]...)
`creators`: EUREC4A community

d70-t:
I like using "EUREC4A community" as author, makes things a lot easier.

I'm a bit hesitant to call the thing "EUREC4A" instead something like "EUREC4A Intake catalogue", because it's a bit unspecific. On the other hand, the goal of this repository is to provide "the EUREC4A dataset", so it's probably the right thing to do?

d70-t:
Do you have any preferences license-wise?
I'd go for `cc-zero`, because I think it's the most useful one for datasets

> It's hard to make reasonable use of datasets without stating where the data is from, so people will do it :+1:. On the other hand, formalizing a requirement to name authors potentially prevents good uses, because depending on the use, it can be hard to e.g. display a predefined statement. Also if we use "EUREC4A community" as author, I guess it's even less useful to go with something like cc-by.

observingClouds:
> I like using "EUREC4A community" as author, makes things a lot easier.
> 
> I'm a bit hesitant to call the thing "EUREC4A" instead something like "EUREC4A Intake catalogue", because it's a bit unspecific. On the other hand, the goal of this repository is to provide "the EUREC4A dataset", so it's probably the right thing to do?

Yeah, this is exactly my thought. Currently "EUREC4A dataset" is maybe a bit vague as the catalog's entries are not immutable but that's where we wanna get to.

observingClouds:
The license is a bit tricky as we reference other data sets that might have specific license. However, for this specific repository I would argue that `cc_zero` is fine, but we should add a note somewhere on how people should add credits to the underlying data and adhere to the [EUREC4A data policy](https://eurec4a.eu/data-simulations/data).

observingClouds (approved):

d70-t:
I just added a note on the data policy and a (very tiny) description of the catalog. Is it still fine?

observingClouds:
Yes, awesome!

d70-t:
Ok, let's see where this will take us...

PR #149, opened 2023-10-09
Update IPFS version

observingClouds:
Bumping IPFS version to add support for IPNI lookups and other enhancements that speed up file lookup and retrieval

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


d70-t:
There's also an IPFS installation in the workflow which runs on pull-requests, coud you update this as well?
Otherwise, this seems to be straight forward.

d70-t (approved):

d70-t:
Unfortunately, the version of the IPFS client is part of our branch protection rules... I just updated the rule to reflect the new version, but probably we'll want to come up with a workflow which doesn't require an update of the protection rules once the client version changed.

PR #150, opened 2023-10-11
find piñata pinned files

observingClouds:
set accelerated DHT client to find datasets faster

### DIFF
--- a/.github/workflows/weekly_tests.yml
+++ b/.github/workflows/weekly_tests.yml
@@ -38,6 +38,7 @@ jobs:
         sudo apt-get update
         sudo apt-get install -y wget curl jq
         bash <(curl https://raw.githubusercontent.com/eurec4a/ipfs_tools/main/install_and_run_ipfs.sh) ${{ matrix.ipfs-version }}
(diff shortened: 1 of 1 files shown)


d70-t:
I'm wondering if this really helps in our situation. As per the [docs](https://github.com/ipfs/kubo/blob/v0.21.0/docs/config.md#routingaccelerateddhtclient):
> Currently, the DHT is not usable for queries for the first 5-10 minutes of operation as the routing table is being prepared. This means operations like searching the DHT for particular peers or content will not work initially. 

So while finding data via DHT should be accelerated in the long term, the initial period after startup is supposed to not be usable (or at least be slow). In the CI setup, we'll (hopefully) be mostly retrieving data during this startup phase...

My guess would be, that if any, [IPNI](https://docs.ipfs.tech/concepts/ipni/) does help in our case.

observingClouds:
I honestly just left this draft open as this was one suggestion I got from the Piñata team when I first pinned the files. However, it currently works very smoothly without (look at all the successful weekly tests 💪 ) so I guess their service is doing something right 🤣 I can close this for now and hopefully we get reminded about this when we face issues again.

PR #152, opened 2023-10-17
add BCO isotope data

observingClouds:


barbados/BCO/bco.yaml
```diff
@@ -83,3 +83,13 @@ sources:
     args:
       urlpath: ipfs://QmeAL3DLEbWS2c2bkyA7mSgopN2ANmtMAGu79mM5GFQGhR
       consolidated: true
+  isotopes:
+    description: Calibrated stable water isotope data in precipitation from the BCO during EUREC4A
+    driver: netcdf
+    args:
+      urlpath: simplecache::https://observations.ipsl.fr/aeris/eurec4a-data/BARBADOS/BCO/precipitation/EUREC4A_BCO_stable_water_isotopes_precipitation.nc
```
RobertPincus, line 90: Seems very straightforward. Can you explain what the simplecache does?

RobertPincus (approved):

barbados/BCO/bco.yaml
d70-t: `simplecache` is a [fsspec cache](https://filesystem-spec.readthedocs.io/en/latest/features.html?highlight=cache#caching-files-locally), without need for user configuration, and which will delete itself on interpreter close.

In this context,
* it caches the **entire** netCDF file locally
* it thereby ensures the file has a path on the local filesystem (which is not the case for `https`, but required for netcdf)

observingClouds: The file is only 20KB in this specific case so caching it locally should not be a problem and more reliable than accessing it through THREDDS 🤣

d70-t: Yes definitely.
I'm sure that this is the right option here.

PR #153, opened 2023-11-14
Add further isotope product

observingClouds:


barbados/BCO/isotopes.yaml
```diff
@@ -0,0 +1,23 @@
...
+sources:
+  precipitation:
+    description: Calibrated stable water isotope data in precipitation from the BCO during EUREC4A
+    driver: netcdf
+    args:
+      urlpath: simplecache::https://observations.ipsl.fr/aeris/eurec4a-data/BARBADOS/BCO/precipitation/EUREC4A_BCO_stable_water_isotopes_precipitation.nc
+      chunks: null
+    metadata:
+      maintainer: Leonie Villiger <leonie.villiger[at]env.ethz.ch>, Franziska Aemisegger <franziska.aemisegger[at]env.ethz.ch>
+      doi: https://doi.org/10.25326/242
```
leonie-villiger, line 14: We measured isotopes in water vapour with two instruments (LGR and Picarro) in parallel. Thus, the second dataset of isotopes in water vapour is missing. 
```suggestion
  water_vapor:
  description: BCO 
  driver: netcdf
  args:
  urlpath: simplecache::https://observations.ipsl.fr/aeris/eurec4a-data/BARBADOS/BCO/PICARRO/EUREC4A_BCO_PICARRO_WVISOS_L1.nc
  chunks: null
  metadata:
  mainainer: Franziska Aemisegger <franziska.aemisegger[at]env.ethz.ch>, Leonie Villiger <leonie.villiger[at]env.ethz.ch>
  doi: https://doi.org/10.25326/245
```

CHANGELOG.md
```diff
@@ -3,7 +3,8 @@
 ## (unreleased)
 
 ### New Datasets
-* Add isotope data from BCO (#152). By [Hauke Schulz](https://github.com/observingClouds)
+* Add precipitation isotope data from BCO (#152). By [Hauke Schulz](https://github.com/observingClouds)
+* Add water vapor isotope data from BCO (#153). By [Hauke Schulz](https://github.com/observingClouds)
```
leonie-villiger, line 7: note sure, whether you need to mention that you added two water vapor isotope datasets

barbados/BCO/isotopes.yaml
observingClouds: Thanks @leonie-villiger that is very helpful to know. How can these two water vapor datasets best be distinguished? What would be a meaningful and general distinct access pattern?
- `cat.BCO.isotopes.water_vapor.CRDS` and `cat.BCO.isotopes.water_vapor.OA_ICOS` (name from Bailey et al.)
- `cat.BCO.isotopes.water_vapor.PICARRO` and `cat.BCO.isotopes.water_vapor.LGR`
- `cat.BCO.isotopes.water_vapor.dry_mole_fraction` and `cat.BCO.isotopes.water_vapor.wet_mole_fraction`

Thinking about the other water vapor datasets on Meteor, ATR, P-3, it seems to me that the type of mole fraction might be the most general identifier that would also work for those datasets. What do you think?

leonie-villiger: I would stick with the naming from Bailey et al., i.e., `*.CRDS` and `.OA_ICOS`, since it is the official EUREC4A paper describing all the isotope datat set.

CHANGELOG.md
observingClouds: No need to :)

observingClouds:
@leonie-villiger I encountered an issue with the PICARRO dataset. The time dimension has an unsupported unit leading to an error in the CI but also when opening the dataset with common tools like panoply as the time dimension cannot be interpreted. Could you change that and release a new version of the dataset?

```
netcdf EUREC4A_BCO_PICARRO_WVISOS_L1.nc {
  dimensions:
    time = 51838;
  variables:
    double time(time=51838);
      :units = "seconds since 19700101_00";
```

`seconds since 1970-01-01 00:00:00` would work for example if the time is given in UTC. Please have a look at the time dimension information given in [CF-Conventions](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.10/cf-conventions.html#time-coordinate) which are widely supported.

observingClouds:
@leonie-villiger I also encountered an issue with the rainwater isotope product. The variable `flag` has the type `long` (int64). This is generally fine but when accessing the file via OPeNDAP this causes an issue because the datatype is not supported. AERIS provides this access to your dataset via this protocol (see Download | Data access section at  https://doi.org/10.25326/242) and we generally try to access these datasets the same way as it allows us to load the datasets lazily, i.e. do not need to download the entire dataset. 

Looking at the dataset it seems the flag has only the values `[0,1,2]` so you could just use `short` (int16) instead. @d70-t has put together a nice overview about problematic data types at https://howto.eurec4a.eu/netcdf_datatypes.html

Could you look into this as well a release a version of the dataset with a different datatype? In this particular case, we might also use a different access protocol to circumvent this issue because the dataset is in total only 21KB 🤣 and downloading the entire dataset does not hurt.

leonie-villiger:
@observingClouds: working on both issues. I will let you know as soon as we succeeded releasing new versions of the datasets

barbados/BCO/isotope-analyzer-wv.yaml
```diff
@@ -0,0 +1,27 @@
+plugins:
+  source:
+    - module: intake_xarray
+
+sources:
+  CRDS:
+    description: BCO CRDS Continuous In-Situ Water Vapor Isotopic Composition
+    driver: opendap
+    args:
+      urlpath: https://observations.ipsl.fr/thredds/dodsC/EUREC4A/BARBADOS/BCO/PICARRO/EUREC4A_BCO_PICARRO_WVISOS_L1.nc
```
leonie-villiger, line 10:
```suggestion
      urlpath: https://observations.ipsl.fr/thredds/dodsC/EUREC4A/BARBADOS/BCO/vapour/EUREC4A_BCO_PICARRO_WVISOS_L3.nc
```
That is the official version (i.e., the one linked from https://doi.org/10.25326/245)

observingClouds: Hi @leonie-villiger, just wanted to check if you had a chance to look into this again 😬

leonie-villiger: - A new version of the BCO precipitation isotope product with corrected datatype of `flag` variable has been uploaded and is now accessible via OPeNDAP. 
- Franziska will take care of the unsupported time dimension in the BCO PICARRO dataset.

Issue #30, opened 2021-01-18
versioning catalog


leifdenby:
to avoid breaking dependent code what do people think about versioning the catalog? It would be nice to be able to refer to a specific git-tagged version when fetching the catalog.

I think with what we've got now we could create a version `v1.0.0`. If we adopt semver (https://semver.org/) we could update the MAJOR version we if we remove/add an endpoint and update the MINOR version when we add new endpoints, and finally update the PATCH version when we make fixed to existing endpoints. Adding/removing endpoint arguments could also be considered a breaking change I think.

Thoughts?

d70-t:
I'll try to write down some of my thoughts on that topic.

# Requirements
* We need a form of stable references to a catalog, as changes will happen, but users certainly want to re-run their old scripts exactly as they are with exactly the original data.
* We need to be able to release updates to the catalog quickly, such that newly available datasets can be used properly without delay or workarounds.
* We want people to use the newest available data by default.

# Semver for the intake catalog
* If a new dataset is added to the catalog, this should be a non-breaking change which adds functionality -> **MINOR** change
  * **except** a user enumerates the catalog and operates on the result of that enumeration, then this could be a breaking change -> **MAJOR** change
* If a dataset is updated, this is quite probably a breaking change for some users (even if it is only correcting a typo) -> **MAJOR** change
* Datasets should not be removed. If they are -> **MAJOR** change
* Datasets should ideally not be renamed, if they are -> **MAJOR** change
* Currently there are substantial changes happening to files on the Aeris server (even if they already have a DOI allocated) these would be a **MAJOR** change for semver but can't be addressed as the catalog is not touched (to be fair, that's an issue which must be solved outside of the catalog)
* I see very little opportunities for **PATCH** updates, but of course, that's not required 🤷 

# Current status
Currently there are some active efforts to fill the catalog with more and more data around HALO, so I'd expect quite a few more changes in the next days. I am also not yet sure if the catalog hierarchy is already settled firmly enough so that renaming isn't necessary anymore, but I hope that we can reach that state soon.

# Thoughts
[...]

PR #155, opened 2023-11-15
Add CITATION.md

observingClouds:
This tries to provide recommendations on how to cite the catalog following the discussion in #30:
- add instructions on how to cite the EUREC4A intake catalog
- add note that datasets within the catalog need to be explicitly referenced as well
- mention that the catalog itself cannot insure integrity of the datasets


leifdenby (approved):
LGTM :)

PR #159, opened 2023-12-15
add ATR turbulence data

observingClouds:
Let us start discussing how to add the ATR turbulence data 🥳


observingClouds:
@pebrilou could you maybe comment on some of the above mentioned points and see if we have used the correct metadata. Feel free to expand the description too.

PR #164, opened 2024-04-03
Update BCO precipitation isotope product

observingClouds:
- `flag` variable has been changed from `long` to `short` to support openDAP protocol
- fix for weekly test as old file is deleted


d70-t (approved):
LGTM :+1:

PR #168, opened 2025-02-16
replace outdated miniconda action setting

observingClouds:
The weekly tests were disabled due to inactivity of the repo. I reenabled the workflow and run into the error that the conda environment was installed in a no longer supported way.

This PR updates the action so we can see [which datasets are currently unreachable/disappeared](https://github.com/eurec4a/eurec4a-intake/actions/runs/13355107684/job/37296756407) 😬


d70-t (approved):
Thanks 🙏

